Issue #6, opened 2015-10-17
Qparameter verbosity


iambernie:
Can we remove the print statements in the `Qparameter` method or at least give the user the option to silence them?

```
>>> from amuse.ic.plummer import new_plummer_model
>>> model = new_plummer_model(1000)
>>> model.Qparameter()
making graph
constructing MST
Q: 1.53074998216
1.5307499821583916
>>> 
```

spzwart:
Dear Bernie,

We will look into this on Tuesday.

Cheers,
Simon

----- Reply message -----

PR #8, opened 2015-10-17
remove print statements in Qparameter

iambernie:
This PR removes 3 print statements in the Qparameter method, see #6


ipelupessy:
it may have been better to have a verbose option (like in connected_components)

iambernie:
I did consider that, but the print statements didn't seem useful/informative enough...

PR #18, opened 2015-11-19
fix for non-numeric version strings. note the comparison may not make…

mherkazandjian:
… sense in this case


ipelupessy:
looks ok

PR #28, opened 2016-01-26
Memof fix

ipelupessy:
fix for fail of:

```
import numpy
from amuse.units import units

n=10000

for i in range(n):
  vec_i = numpy.random.random(size=[3]) | units.none
  print i, vec_i, vec_i.lengths(), vec_i.lengths().value_in(units.none)
```
another way to fix this would be to replace the various instances of none_unit('none','none') in e.g. to_simple_form in units/core.py...

### DIFF
--- a/src/amuse/test/amusetest.py
+++ b/src/amuse/test/amusetest.py
@@ -8,7 +8,6 @@ import inspect
 from amuse.support import exceptions
 from amuse.support import literature
 from amuse.support import options
-from amuse.units.si import no_unit
 from amuse.units.quantities import Quantity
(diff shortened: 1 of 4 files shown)


iambernie:
This reply won't help the issue a single bit, but if you're only interested in the magnitude of the vectors in `units.none`, this will workaround the problem:

```python
import numpy
from amuse.units import units

n = 10000
vectors = numpy.random.random(size=(n, 3))
vec_lengths = numpy.linalg.norm(vectors, axis=1) |units.none
```

jilkoval:
The fix is working for my problem. Thanks.

PR #39, opened 2016-02-14
fix for N+1 message length problem

### LINKED ISSUE #38: Exception when calling with length = max_message_length+1
the following bombs:
```
N=1001
p=new_plummer_gas_model(N)
f=Fi(max_message_length=1000)
f.gas_particles.add_particles(p)
```
while N=1000 or N=1002 doesn't...

ipelupessy:
this seems to fix #38, I think it was assumed call_count was evaluated at function def time, but it evaluated when the function is callend, call_count is changed in the loop.

however there may be deeper problems with the way split_input_array is used...and it does not look particularly efficient for call_counts=1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -221,6 +221,8 @@ src/amuse/community/mocassin/worker_code.f90
 src/amuse/community/mpiamrvac/worker_code.f90
 src/amuse/community/nbody6xx/worker_code.f90
 src/amuse/community/nbody6xx/nbody6xx_worker
+src/amuse/community/pikachu/worker_code.cc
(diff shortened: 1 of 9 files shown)


ipelupessy:
also this needs a test, but I have not yet reproduced it independent of Fi (the bug is not fi related)

ipelupessy:
I have simplified some more (I think call_count=1 is actually an error), but some of it crossed with 994a593, I think it merged ok, but please double check..

ipelupessy:
it could be I am misunderstanding the reason for the call_count=1 case, and oversimplifying. tests are ok here I think, so if that is the case a test case is needed...

PR #54, opened 2016-04-06
Rebound changes: add support for softening and for setting boundary types (only for some integrators)

rieder:


ipelupessy:
do you know what the end_time_accuracy_factor in the test_rebound.py is??

rieder:
Not sure - could it be related to exact_finish_time ("set to 1 to finish the integration exactly at model_time. Set to 0 to finish at the next dt")? That one is currently hard-set to 1, but I changed it in my working branch to a proper parameter.

ipelupessy:
I think if it is possible to finish at the exact time this is preferable (spec of AMUSE is to finish at end_time or later), so I don't think this parameter is necessary..but then the paramter exact_finish_time should always be 1 (and end_time_accuracy_factor removed from the test)...

rieder:
Well, I added it for testing purposes (was getting weird timestep problems, but I think Arjen solved that problem now). It's an optional parameter in Rebound itself. I agree, end_time_accuracy_factor seems to have purpose to me and should be removed if that is indeed so.

ipelupessy:
ok, i have removed end_time_accuracy_factor from the test (just to double check: it was not defined in interface.py of master)

arjenve:
It's a parameter in Hermite (the test may have come from that), it is used to stop the simulation just before, exactly on or just after the given time in evolve_model. (it is now only in Hermite, as an "experimental" feature).

PR #63, opened 2016-06-08
adding a galactic potential into community folder. The examples are in the ext/ folder

anamabo:


anamabo:
Interface of galactic_model added to /src/amuse/community
examples of galactic_model added to /src/amuse/ext

ipelupessy:
there are are still a few issues:
- binary files (from the operating system?) should not be part of the commit
- examples should be in the examples directory, not in ext

anamabo:
I moved the examples of galactic model to the folder examples.
I could not remove hidden files created by MAC.

ipelupessy:
it ok, the hidden/system files have been deleted (I think they were only present in the ext)
minor things:
- maybe you can give it a more specific name than "galactic_model"
- maybe you can put/move the references in the docstring of the galactic model interface so you get references at the end when you run a script (look at other codes for the format)
- I think the manual you could put into the doc dir somewhere (maybe we should ask arjen where)??

anamabo:
Hi Inti:

Regarding to your previous comments:
1. This interface contains several analytical models for the Galaxy and maybe it will be expanded in a future, so that's why I named it as 'galactic_model' 
2. I put the manual together with the examples just to make easier the comprehension of the scripts. 

Is it fine if I leave the branch in the way it is now?

ipelupessy:
1) there might be more than one galactic model code(like there are more than 1 gravity codes), so a distinguishing name would be nice in community, even if something like initials+year galactic_model_cm2014 - it may make sense to put a galactic_model module in the amuse/ic (something Lucie more or less started), but this would be a wrapper pointing to the respective codes...
2) ok, no problem (we can always move it..)

anamabo:
Can I name the interface as "galaxia" or "galaxy"

ipelupessy:
"Galaxia" would be nice (or maybe ViaLactea :-) )

anamabo:
Hi Inti: I've added galaxia into the community folder and examples_galaxia into the examples folder. Please, let me know if everything is fine, so I can let people know that they are ready to use galaxia!

Thanks!
Carmen

PR #65, opened 2016-08-24
add array vector parameters 

ipelupessy:
(can potentially be variable size)


ipelupessy:
proposal; I may need to test it a little bit more

PR #68, opened 2016-09-08
Allowed evolution to and synchronization at a specified time (not nec…

stevemcmillan:
…essarily

a natural time step).


ipelupessy:
is it possible to add a test case

stevemcmillan:
Script test_sync.py illustrates use. I implemented force_sync as a parameter, mainly because I don't know how to add an argument to evolve_model(). If someone can do it better, please do.

ipelupessy:
force_sync as parameter is ok (actually preferred)

stevemcmillan:
Has this been merged into amusebase yet? Josh needs to use it with Flash.

PR #69, opened 2016-09-11
Two small bug fixes

rknop:
hacs64 wouldn't compile on Ubuntu 16.04 without -D_FORCE_INLINES

There was a memory handling bug in the bonsai interface.

### DIFF
--- a/src/amuse/couple/multiples.py
+++ b/src/amuse/couple/multiples.py
@@ -1781,7 +1781,7 @@ def find_nn3(plist, field, G):
         phimin = 0.0 | nbody_system.energy
         dxmin = 1.e30
         pmin = None
-        j = numpy.argmin(phi)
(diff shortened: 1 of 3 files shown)


rknop:
Added a third small bugfix in multiples.py

src/amuse/couple/multiples.py
```diff
@@ -1781,7 +1781,7 @@ def find_nn3(plist, field, G):
         phimin = 0.0 | nbody_system.energy
         dxmin = 1.e30
         pmin = None
-        j = numpy.argmin(phi)
+        j = numpy.argmin(phi.number)
```
ipelupessy: the fix is ok; did you have it failing? it works here...
( a safe alternative is j=phi.argmin())

PR #71, opened 2016-10-31
adding orbital_elements_for_rel_posvel_arrays and tests

jilkoval:
Adding routine (and tests) to calculate orbital elements for number of two-body problems. This can be useful for number of orbiters of the same central body. Arguments are arrays of relative position and velocity vectors, and vector of total masses of the individual two-body problems. The routine should be reasonably fast even for a large numbers of systems (10^6).

### DIFF
--- a/src/amuse/ext/orbital_elements.py
+++ b/src/amuse/ext/orbital_elements.py
@@ -179,7 +179,7 @@ def orbital_elements_from_binary( binary, G=nbody_system.G):
         
     return mass1, mass2, semimajor_axis, eccentricity, true_anomaly, inclination, long_asc_node, arg_per
 
-def orbital_elements_for_rel_posvel_arrays(rel_position, rel_velocity, total_masses, G=nbody_system.G):
+def orbital_elements_for_rel_posvel_arrays(rel_position_raw, rel_velocity_raw, total_masses, G=nbody_system.G):
     """
     Orbital elements from array of relative positions and velocities vectors,
     based on orbital_elements_from_binary and adapted to work for arrays (each line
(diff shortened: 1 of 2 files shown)


ipelupessy:
have to look at it in more detail yet, but one thing that isprobably not correct/consistent are the 
   .as_quantity_in(units.yr)
(or AU etc)
because these will not work with nbody_system.G input (the default).
there is probably no reason to go to yr or make any ref to units  anyway...

jilkoval:
I see, that makes sense. I think .as_quantity_in(...) can indeed be easily omitted. However, I use the reference to units also in the auxiliary routine normalize_vector(vecs, norm, vecs_unit, one_dim = False), which normalizes multiple vectors that are in an array. In detail, I specify the shape of the output array as:
vecs_norm = numpy.zeros_like(numpy.array(vecs.value_in(vecs_unit)))
Basically, I just need to strip the units, since the result is unit-less. Is there a better way to do this?

ipelupessy:
I think:
numpy.zeros_like(vec)
will work...

jilkoval:
Actually, there is a catch with this (and this is probably why i used value_in in the above, despite I never reported the trouble). Check this example:
```
import numpy
from amuse.units import units
v1 = numpy.random.random(3) | units.m
v2 = numpy.ones_like(v1)

# the following gives an error
print numpy.sqrt(v2)

# I think it is because the dtype of v2 is 'object'
print v2.dtype

# possible fix
v2 = numpy.ones_like(v1, dtype=numpy.float)
print numpy.sqrt(v2)
```
Is the fact that the new array created via numpy.zeros_like (or similar) is an object all right or should it rather be a numerical type?

In any case, I will add the dtype in the routine, and I think it should work.

ipelupessy:
I see, you are right, in that case I would prefer something like :
numpy.zeros(vec.shape)
if that works...(ones_like doesn't take dtype here)

jilkoval:
ok, should be fixed now

ipelupessy:
the test is not working here, because orbital_elements_for_rel_posvel_arrays is not imported...
also:
for me (poor soul still on numpy 1.5.1) einsum is not working...is it easy to work around this?

ipelupessy:
I think e.g.
(e_vecs_unit[filter_non0_ecc]*asc_node_matrix_unit[[filter_non0_ecc]).sum(axis=-1)
should do

jilkoval:
Indeed, I think I used this in my previous version too. However, I switched to einsum, because it is much faster for big N. I will come back with a test.

Sorry about the missing import in the test, will fix that.

ipelupessy:
ok, great!
also: there is also a bunch of unit.none things which I don't think are necessary, and also:
  mom[:,2]<0.|(units.m*units.kms)
can be replaced with
  mom[:,2]<0.*mom[0,0]
 or something (to be unit system agnostic)
(did you think about the N=1 case proposal?)

jilkoval:
* I removed units.none (only one is left, which I think does make things simpler). Mostly I used units.none so i could use some amuse linear algebra routines (like lengths(), cross, dot) so I replaced those with their numpy equivalents.
* I replaced the numpy.einsum by your suggestion, which is only a little bit slower (0.00559s vs. 0.00533 for N=10^3 and 0.368s vs 0.347s for N=10^5). Just for completeness, what I was using initially was actually numpy.diag(e_vecs_unit[filter_non0_ecc].dot(asc_node_matrix_unit[filter_non0_ecc].T)) which is indeed much slower (0.147 for N=10^3 !).

I will look at the N=1 case now.

jilkoval:
I introduce an array dimension handling at the begging of the routine (it expands the dimension of the array in case N=1). It is indeed much cleaner now.

ipelupessy:
ok, merged now ..

PR #77, opened 2017-01-04
Add REarth and MEarth as standard units

rieder:
Earth mass and radius are practical units for planetary simulations, useful to have them pre-defined.


ipelupessy:
I may change the Rearth value slightly to 6371.0088 at some point (IUGG mean radius)

rieder:
Thanks for merging!
Best not wait with changing the value to IUGG mean radius then, otherwise re-running a simulation using this value will have changed results...

ipelupessy:
good point! done..

PR #81, opened 2017-01-17
Update to newer Rebound version, with new integrators

rieder:
(and old ones removed)

Update rebound to use git checkout 7184fb43852a6c3721f87dc8b5f6878a540a8dd0.

This version works, though test6 fails in my setup (see below). The result is the same over multiple tries.

rieder@Usajii Amuse $ nosetests-2.7 test/codes_tests/test_rebound.py
.....F............
======================================================================
FAIL: test6 (test.codes_tests.test_rebound.ReboundInterfaceTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/rieder/Documents/Code/Amuse/test/codes_tests/test_rebound.py", line 135, in test6
    self.assertEquals([0, -10], instance.new_particle(0.5,  0.5, 0, 0,  0, 0.5, 0, 0.01, 1).values())
  File "/Users/rieder/Documents/Code/Amuse/src/amuse/test/amusetest.py", line 77, in failUnlessEqual
    self._raise_exceptions_if_any(failures, first, second, '{0} != {1}', msg)
  File "/Users/rieder/Documents/Code/Amuse/src/amuse/test/amusetest.py", line 64, in _raise_exceptions_if_any
    raise self.failureException(msg or err_fmt_string.format(first, second, *args))
AssertionError: [0, -10] != [1094929759, -10]

----------------------------------------------------------------------
Ran 18 tests in 2.043s

FAILED (failures=1)


ipelupessy:
I think the test result was wrong (or worked by coincidence): the tests checks whether an error
(-10) is returned, the 1st value is the particle index, but undefined in the code.... 

I am still having trouble with building (patches not working)

ipelupessy:
ok, that's just me having trouble with empty patch sets? I can build by retrying...

rieder:
I figured it might be something like that.

Actually, there are no patches (I didn't change that from the previous version): none are necessary, save maybe one change to the Rebound makefile (-march doesn't seem to work on macOS).

rieder:
I had that problem with the patches - but it disappeared after retrying. Guess that wasn't just me.

ipelupessy:
for me its ok now...

rieder:
Thanks for merging!

ipelupessy:
the march thing arjen thinks he had fixed....will be later maybe

PR #86, opened 2017-01-28
 fi minor change, move abs acc to seperate array

ipelupessy:
This is mainly better style, but could also help in parallel perf.


rieder:
... ready to merge?

ipelupessy:
reminder to self: needs bitwise comparison test for final approval..

ipelupessy:
close because in favour of #554

PR #87, opened 2017-01-28
Fi minor init fix

ipelupessy:
this speeds up init by putting the particles in morton order right away.


rieder:
Any reason not to merge this?

PR #89, opened 2017-02-01
Change all tabs to four spaces, to get rid of mixed tab/space indenting

rieder:
This makes the interface much more readable if a tab is not four spaces in your editor...


ipelupessy:
good idea; tabs should not be used...

rieder:
Note that this was only done in the interface.cc of Rebound, for now.

PR #92, opened 2017-02-03
Mercury initial integrator options

jilkoval:
* **mercury_main.for** -- fix of the documentation to correspond to the actual code (and documentation elsewhere in Mercury):
[main code](https://github.com/amusecode/amuse/blob/master/src/amuse/community/mercury/src/mercury6_2.for#L4971) integration is stopped _if opt(1).eq.1_
[manual](https://github.com/amusecode/amuse/blob/9966274f6476da14f4e63c35a5549b491803f0e7/src/amuse/community/mercury/src/mercury6.man#L403) _Should the integrator stop if a close encounter occurs (yes or no)?_ with the classical 1=yes, 0=no
* **amuse_helpers.f90** -- default integration options to match [param.in](https://github.com/amusecode/amuse/blob/master/src/amuse/community/mercury/src/param.in#L13)


ipelupessy:
just to be sure: the only change affecting the interface is that it now should continue when a close encounter occurs? Have you tested this?

jilkoval:
Indeed, the only effective change is the `opt(1)=0` (instead of the original `opt(1)=1`). The rest should only affect the output format of Mercury.

I made some tests and it seems that the integration does not stop in either case. The only difference I can find is that in case `opt(1)=1` (i.e. when the integration should stop), Mercury outputs info on the close encounters in the _information file_ (`outfile(3)`); a message similar to:
```
WARNING:  Stopping integration due to an encounter between         4,      16
         at           0.114years 
```
for (all?) occured close encounters. Btw, It seems that this corresponds to messages _encounter minima occured_ from the redirection file. But the integration does not stop and the results for the particles that experienced close encounters do not seem to be affected either. 

After looking into **mercury6_2.for** and **amuse_helpers.f90**, I'm confused. Could it be related to the comment [on line 897 in amuse_helpers.f90](https://github.com/amusecode/amuse/blob/master/src/amuse/community/mercury/src/amuse_helpers.f90#L897)?

ipelupessy:
about 4b221fb : I would be hesitant with such  part of code being turned off (especially in mercury), however since there is a comment about it it should be ok..nonetheless maybe its an idea to make it settable from the interface with a parameter and making this piece conditional on that?

ipelupessy:
also let me know when you think you have really diverged too much from what good behaviour for mercury is (or when it breaks something)..also I think its easier if we just commit what are improvements anyway (and do this quickly)??

jilkoval:
I agree that [4b221fb](https://github.com/amusecode/amuse/commit/4b221fb17f84d7ef84f1e7baf1f14ca5abcb608e) is not ideal solution at all. I introduced it to get some tests running. I think it would be ideal to implement removing particles in Mercury and also make it possible to turn it off from the amuse interface. Also note that correctly, the same changes need to be introduced in the variable-timestep integrator routine `mal_hvar` ([here] (https://github.com/jilkoval/amuse/blob/mercury/src/amuse/community/mercury/src/amuse_helpers.f90#L1294)). Similarly, the removal of particles by Mercury in collisions should be fixed. 

My Mercury branch is quite experimental at the moment, to make some tests running. Hopefully and ideally, I can solve things properly soon ;)

rieder:
What's the status of this? Ready for merging?

ipelupessy:
its ready exept for one thing: the collision setting and the bit about the commented out code.

In principle the changes should disable the mercury internal collision detection both for orbiters (the OPT(2) setting and central body (commented out bit). This is necessary because otherwise the amuse state becomes inconsistent... ideally this should be tested.

ipelupessy:
just to be on the safe side, creating new PR

PR #104, opened 2017-03-10
Update lab.py to also import Rebound interface

rieder:
I think Rebound is by now tested well enough to be included as a default interface.


ipelupessy:
ok..although I consider importing from lab bad form (don't do it for anything you may pass on to other people, including your future self;-) )

rieder:
Agreed - maybe lab.py should be deprecated? And probably removed with a "lean" (A)MUSE...

Issue #105, opened 2017-03-10
Memory leak in Rebound


rieder:
Memory usage in Rebound seems to increase over time for my debris disc simulations. I'm not yet sure what causes this, I'll investigate. 
Possibly it could be related to synchronising back and forth with Amuse?

rieder:
I'd assign myself to this issue but it seems I can't do that.

rieder:
The memory leak occurs when deleting particles in Rebound (tested using script below).

`
import numpy as np
from amuse.lab import *
from amuse.community.rebound.interface import Rebound

timestep = 0.01|nbody_system.time
particles = new_plummer_sphere(1000)

gravity = Rebound()

gravity.parameters.integrator = "whfast"
gravity.parameters.timestep = timestep

gravity.particles.add_particles(particles)

for i in range(1000):
    gravity.evolve_model((0.5 + i)*timestep)
    p = particles[0:1]
    gravity.particles.remove_particles(p)
    particles.remove_particles(p)
`

rieder:
On a related note, deleting particles in Rebound is very slow. 
A workaround for this problem is to simply set the mass of the to-be-deleted particle to 0, that way there is no memory leak, the code runs faster and the particle remains as a "tracer" particle only.

rieder:
The problem seems to be in the method used to delete particles from a running Rebound simulation. It looks like when this happens, the whole particle set is re-initialised minus the removed particles (reb_remove_all). I am not sure why this is done, wouldn't it be better to remove only those particles (with a reb_remove() or reb_remove_by_hash() call)? 
CC @arjenve

rieder:
I have a preliminary fix, which uses reb_remove_by_hash() to delete a particle. Seems to work, and should be much faster.

PR #108, opened 2017-03-13
Patch rebound delete

rieder:
Fixes for bug #105. Only still fails on the final part of test3, can't figure that out yet (help would be appreciated). But deleting particles this way seems to work correctly, and much faster than previous method.

### DIFF
--- a/src/amuse/community/rebound/interface.cc
+++ b/src/amuse/community/rebound/interface.cc
@@ -467,17 +467,13 @@ int get_index_of_next_particle(int index_of_the_particle,
     return 0;
 }
 
-int delete_particle(int index_of_the_particle){
-    particle_location loc = get_index_from_identity(index_of_the_particle);
-    if(loc.index < 0) {return -1;}
-    
-    indexMap.erase(index_of_the_particle);
-    for( ReboundSimulationVector::iterator i = codes.begin(); i != codes.end(); i++) {
-        code_state cs = *i;
-        if(cs.code == loc.code){
-            cs.has_removal = true;
-        }
-        *i = cs;
+int _delete_particle(int index_of_the_particle, int code_index){
+
+    if(code_index < 0 || code_index >= (signed) codes.size()){
+        return -10;
+    }
+    int keepSorted = 1;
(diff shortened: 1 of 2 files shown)


rieder:
(more testing needed)

rieder:
All tests run without problems, but I need to re-add the detection of massless particles after the initial commit.

ipelupessy:
should it wait for that (detection of massless particles after the initial commit)?

arjenve:
I know it is being used (but maybe not unit tested). So it is essential....

rieder:
I think fixing this memory leak is more important than support for non-active particles. 
But please wait a little while longer, I'm working on it this week so I hope to get it done soon. 
If I can't get it done this week, please merge this one first and I'll work on re-adding massless particle resorting afterwards.

rieder:
@arjenve am I right in thinking that in the ordering of particles the only important issue is that massless particles should be last (and that N_active == number of non-massless particles)?

arjenve:
Yes correct, this is the only reason we need to sort....

rieder:
Ok, great! Thanks!

ipelupessy:
have you tested this last bit (the massless particles) , not sure it is covered by the tests...

rieder:
It's not covered by the tests, I'm not sure how to test it in a "nosetests" way. (This wasn't covered by tests before either...)
However, I have done this manually, by manually printing N_active value and particle attributes after changing particle masses, and by comparing execution time between a simulation with mostly massless particles and one with the same N but no massless particles.

rieder:
This fixes #105

rieder:
Hi,

Is there a way to prevent particles from ending up more than once in a particleset?

Currently, I get this:
p = Particle()
ps = Particles()
ps.add_particle(p)
ps.add_particle(p)
print ps
                 key
                   -
====================
11252531812499082495
11252531812499082495
====================

and I hope to get a set where each key is unique...

Steven

spzwart:
what about the following:

for pi in p:
    if pi not in ps:
         ps.add_particle(pi)

Simon

rieder:
Thanks Simon, that would work.

Also, oops, this should have been sent to the Amuse list, not to Github...

PR #111, opened 2017-03-22
Update Rebound to v3.2.4

rieder:
No problems with tests, and no interface changes.
Changes in this version are in the WHFast-Helio integrator, which is also used by the Hermes integrator.


ipelupessy:
one minor issue - that has nothing to do with rebound, is that the new version will not be used automatically (needs distclean in the rebound dir I think)

rieder:
Indeed. But I tested with a clean Amuse, so that shouldn't be a problem.

PR #113, opened 2017-03-30
Update Rebound interface to not set unused particle attributes

rieder:
Moreover, setting some of these cause problems later on with the Tree gravity solver...


ipelupessy:
if feels dangerous - do the attributes have well defined value if not set?
(i.e. isn't the value given in new_particle just wrong for these attributes?)

rieder:
Rebound doesn't manually set these values by default, so I don't think that's a problem. They are initialised.

PR #125, opened 2017-04-26
Update rebound to latest version, change defaults

rieder:
- Adds support for Janus integrator
- Change default integrator of Rebound interface to IAS15, same as Rebound standalone default.


rieder:
Janus uses different (integer) arrays for storing positions and velocities internally, so changing velocities/positions will require setting  
`code->ri_janus.recalculate_integer_coordinates_this_timestep = 1`
for it to work.

PR #129, opened 2017-05-11
brutus checkin 

ipelupessy:
matches old svn version


ipelupessy:
needs to be checked by Tjarda if it is good enough...

ipelupessy:
superseded by #130

PR #133, opened 2017-05-29
add f90nml to install.py

ipelupessy:


ipelupessy:
this is a light library useful for fortran codes that work with namelists (such as encountered in OMUSE); it may be useful to include this (currently only IFS on the meteo branch of OMUSE requires it)

rieder:
Maybe it's better to have install.py be branch-dependent? I'd rather not install things that aren't used...

ipelupessy:
you are right...in any case the prereq lib install scripts (there are are 3 now) can/should be merged (and then you would automatically want different options for the different cases standard/ bin distribution/ omuse/ etc)

ipelupessy:
closing for the moment; better done in the context of refactoring installation scripts and or generalizing the framework, with domain specific packages... AMUSE/OMUSE/HyMUSE etc)

PR #136, opened 2017-06-06
Add units to angles and vectorise 'orbital_elements.py'

rieder:
- Fix old to-do: angles now use AMUSE units
- Use trigo functions from Amuse to make angles unit-aware

Possible danger: previous default for angles was in degrees. However, unitless angles will now be seen as radians.
Possible workaround: detect unitless angles somehow, and assign units.deg to those? Is this possible/desired?

Also, some code cleanup (indents to multiples of 4, get rid of too long lines) for improved readability

### DIFF
--- a/src/amuse/ext/orbital_elements.py
+++ b/src/amuse/ext/orbital_elements.py
@@ -388,90 +388,9 @@ def orbital_elements_from_binaries(primaries, secondaries, G=nbody_system.G):
     mass1 = primaries.mass
     mass2 = secondaries.mass
     total_mass = mass1 + mass2
-
-    specific_energy = (
-            (1.0/2.0)*velocity.lengths_squared()
-            - G*total_mass/position.lengths()
-            )
-    specific_angular_momentum = position.cross(velocity)
-    specific_angular_momentum_norm = specific_angular_momentum.lengths()
-    specific_angular_momentum_unit = (
-            specific_angular_momentum
-            / specific_angular_momentum_norm
-            )
-
-    semimajor_axis = -G*total_mass/(2.0*specific_energy)
-
-    eccentricity_argument = (
-            2.0*specific_angular_momentum_norm**2
-            * specific_energy
-            / (G**2*total_mass**2)
-            )
-    if (eccentricity_argument <= -1):
-        eccentricity = 0.0
-    else:
-        eccentricity = numpy.sqrt(1.0 + eccentricity_argument)
-
-    # Orbital inclination
-    inclination = arccos(
-            specific_angular_momentum.z
-            / specific_angular_momentum_norm
-            )
-
-    # Longitude of ascending nodes, with reference direction along x-axis
-    z_vector = [0., 0., 1.] | units.none
-    ascending_node_vector = z_vector.cross(specific_angular_momentum)
-    if ascending_node_vector.lengths().number == 0:
-        ascending_node_vector_unit = numpy.array([1., 0., 0.])
-    else:
-        ascending_node_vector_unit = (
-                ascending_node_vector
-                / ascending_node_vector.lengths()
-                )
-    long_asc_node = arctan2(
-            ascending_node_vector_unit[:, 1],
-            ascending_node_vector_unit[:, 0]
-            )
-
-    # Argument of periapsis and true anomaly, using eccentricity a.k.a.
-    # Laplace-Runge-Lenz vector
-    mu = G*total_mass  # Argument of pericenter
-    position_unit = position/position.lengths()
-    e_vector = (
-            (1.0/mu)*velocity.cross(specific_angular_momentum)
-            - position_unit
-            ) | units.none
-    if (e_vector.lengths() == 0.0):
-        # Argument of pericenter and true anomaly cannot be determined for e =
-        # 0, in this case return 1.0 for the cosines
-        cos_arg_per = 1.0
-        arg_per = 0.
-        cos_true_anomaly = 1.0
-        true_anomaly = 0.
-    else:
-        e_vector_unit = e_vector/e_vector.lengths()
-        print(e_vector_unit, ascending_node_vector_unit)
-
-        cos_arg_per = numpy.dot(e_vector_unit, ascending_node_vector_unit.T)
-        # cos_arg_per = e_vector_unit.dot(ascending_node_vector_unit)
-        e_cross_an = numpy.cross(e_vector_unit, ascending_node_vector_unit)
-        # print(specific_angular_momentum_unit, e_cross_an)
-        ss = -numpy.sign(
-                numpy.dot(specific_angular_momentum_unit, e_cross_an.T)
-                )
-        # ss=-numpy.sign(specific_angular_momentum_unit.dot(e_cross_an))
-        sin_arg_per = ss*(e_cross_an**2).sum()**0.5
-        arg_per = arctan2(sin_arg_per, cos_arg_per)
-
-        cos_true_anomaly = numpy.dot(e_vector_unit, position_unit.T)
-        # cos_true_anomaly = e_vector_unit.dot(position_unit)
-        e_cross_pos = numpy.cross(e_vector_unit, position_unit)
-        ss = numpy.sign(
-                numpy.dot(specific_angular_momentum_unit, e_cross_pos.T)
-                )
-        # ss=numpy.sign(specific_angular_momentum_unit.dot(e_cross_pos))
-        sin_true_anomaly = ss * (e_cross_pos**2).sum()**0.5
-        true_anomaly = arctan2(sin_true_anomaly, cos_true_anomaly)
+    semimajor_axis, eccentricity, true_anomaly, inclination, long_asc_node, \
+        arg_per = orbital_elements_for_rel_posvel_arrays(
+            position, velocity, total_mass, G=G)
 
     return (
             mass1, mass2, semimajor_axis, eccentricity, true_anomaly,
--- a/src/amuse/ext/orbital_elements.py
+++ b/src/amuse/ext/orbital_elements.py
@@ -1,3 +1,5 @@
+from __future__ import print_function
+
 import numpy
 
 import warnings
@@ -11,7 +13,7 @@ from amuse.units.quantities import to_quantity, VectorQuantity
 
 def newton(f, x0, fprime=None, args=(), tol=1.48e-8, maxiter=50):
     if fprime is None:
-        print "provide fprime"
+        print("provide fprime")
         return x0
     i = 0
     x = x0
@@ -50,10 +52,10 @@ def equal_length_array_or_scalar(
             if mode == "warn":
                 warnings.warn("Length of array is not equal to %i. Using only\
                         the first value." % length)
+                return array[0:1]
             elif mode == "exception":
                 raise Exception("Length of array is not equal to %i. This is\
                 not supported." % length)
-            return array[0]
     except:
         if mode == "warn":
             warnings.warn("Not an array, continuing with scalar.")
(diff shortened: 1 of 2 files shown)


ipelupessy:
another thing: I wonder if there is a vectorised routine for generating a particle set from arrays of orbital elements (maybe lucie has one?)

jilkoval:
@ipelupessy I agree that a vectorised generation of positions and velocities from elements would be good to have. I used a loop over the orbital elements and get the position and velocity vectors from Kepler (see [relative_position_and_velocity_from_orbital_elements](https://github.com/amusecode/amuse/blob/master/src/amuse/ic/isotropic_cloud.py#L37) in isotropic_cloud.py).

rieder:
Thanks for the feedback. Renaming the functions would be fine. The old one can then use the new one internally, with a conversion to degrees for the angles.

I started working on a vectorised version of the binary generator - might be able to get it done tomorrow.

rieder:
I've added a function that checks if an angle has a unit or not, and adds degrees if this is not the case. This is still somewhat risky, but the advantage is that now the function works as expected for legacy scripts and for scripts that use units for angles. The deprecation warning could be added to the test function, to warn against unitless usage. @ipelupessy can you add such a warning?

rieder:
Regarding vectorising: this is not so difficult in principle, but there are a few issues to consider. Most importantly:
Currently (1), the output is a particleset with the two binary stars. Should it be like this for the vectorised version (e.g. an array of particlesets)? This doesn't seem very useful, as it would still require a loop at some point when using the output.
Or (2) should it be a set of primaries and a set of secondaries? These could be used directly, or they could be used to generate binary pairs.
Or (3) should should it already be a set of stars and a set of binary pairs (as needed for SeBa/BSE etc)? 
The first option would be the least obtrusive, while the last two seem more useful (but they would break compatibility). I would prefer option 2, since there already is a function to convert that to option 3 and option 1 seems not truly vectorised... Maybe renaming the function is necessary after all.

rieder:
By the way, is there a way to detect if the unit system is using Hénon units, SI units or generic units? It would be great if we can get rid of the "G=nbody_system.G" kwarg.

rieder:
I've implemented option 2 for the generation of binaries. Should work as-is, but could use some testing.
Still need to do the reverse: vectorise generating orbital elements from binaries / list of positions and velocities.

jilkoval:
Cool! 
Just few notes: You use true anomaly as one of the input vectors. If mean anomaly is used instead (there might be cases when that is needed) one still has to go through a loop of Kepler. Also, one should be careful when using the routine to generate binaries with the same primary (e.g. particles in a disk), since each binary has different center of mass.

rieder:
Yeah - still thinking about a few aspects of this. 
There seems to be a "true_anomaly_from_eccentric_anomaly" function already - would this help?
Maybe it's best to split things up in a "orbital_elements_to_cartesian" function (which would return m1, m2, relative position and relative velocity) and a "cartesian_to_binary" function (which would return binaries with center-of-mass at (0,0,0)?

jilkoval:
I think that you cannot get true or eccentric anomaly from the mean anomaly without solving Kepler's equation. As far as I know, there is no vectorized version of that in amuse (kepler_orbiters would be one of the places to look, but I didn't find it there). 
I think your function is definitely useful for many cases, so I wouldn't make it more complicated. In the case the central body is the same for all binaries, there are different solutions depending on the problem (if the orbiters are zero-mass, your function can be still used; or otherwise some democratic or Jacobi coordinates might be the way, but that is beyond the scope of this function, I think).

rieder:
Yeah, I think that would be beyond the scope of this function... Nonetheless, splitting it up would seem to be useful and quite straightforward.
Anyway, the function specifically asks for the (kwarg) true_anomaly, so I doubt there will confusion about that bit.

ipelupessy:
this now contains #137 ??

rieder:
@ipelupessy yes. #137 was a mistake.

ipelupessy:
btw for me the tests in test/ext_tests/test_orbital_elements.py break...

rieder:
Working on it ;).

ipelupessy:
I would move ```angle_with_unit``` to inside ```new_binary_from_orbital_elements```Its not something that should be used for anything anywhere else... (unitless numbers should always be interpreted as radians since an angle unit can always be dropped by a numerical operation; the angle_with_unit here sort of an exception for legacy reasons, moving it inside the function will prevent people from importing it!)

rieder:
You're right, I'll do that.

ipelupessy:
maybe its just me who is confused;-) but: what is your plan with the "Still need to do the reverse: vectorise generating orbital elements" mentioned above?? (this is what orbital_elements_for_rel_posvel_arrays does, or are you talking about wrapping this one for various input options??)

rieder:
Just checking if every function is vectorised. You're right, that one works for a number of binaries (or others) already.

src/amuse/ext/orbital_elements.py
```diff
@@ -27,81 +28,228 @@ def newton(f,x0,fprime=None,args=(),tol=1.48e-8,maxiter=50):
...
+            sin_long_asc_nodes*cos_arg_per
+            + cos_long_asc_nodes*sin_arg_per*cos_inclination
+            )
     alphaz = sin_arg_per*sin_inclination
-    alpha = numpy.array([alphax,alphay,alphaz])
-
-    ### beta is a unit vector perpendicular to alpha and the orbital angular momentum vector ###
-    betax = -cos_long_asc_nodes*sin_arg_per - sin_long_asc_nodes*cos_arg_per*cos_inclination
-    betay = -sin_long_asc_nodes*sin_arg_per + cos_long_asc_nodes*cos_arg_per*cos_inclination
+    alpha = numpy.array([alphax, alphay, alphaz])
```
ipelupessy: one thing to watch for is whether numpy arrays of python objects (ie quantities) are generated (this will slow down things tremendously)  - this may do that  (but maybe not)

src/amuse/ext/orbital_elements.py
```diff
@@ -27,81 +28,228 @@ def newton(f,x0,fprime=None,args=(),tol=1.48e-8,maxiter=50):
...
+        angle = angle | default_unit
+    return angle
 
-    cos_long_asc_nodes = numpy.cos(longitude_of_the_ascending_node)
-    sin_long_asc_nodes = numpy.sin(longitude_of_the_ascending_node)
 
-    ### alpha is a unit vector directed along the line of node ###
-    alphax = cos_long_asc_nodes*cos_arg_per - sin_long_asc_nodes*sin_arg_per*cos_inclination
-    alphay = sin_long_asc_nodes*cos_arg_per + cos_long_asc_nodes*sin_arg_per*cos_inclination
+def equal_length_array_or_scalar(
```
ipelupessy: not sure i like this one - or at least its default behaviour (ie taking the first value if the lengths are unequal). if the length are unequal an exception should be raised (but then that is the default behaviour...)...but maybe I am not seeing a particular case where this is handy...

src/amuse/ext/orbital_elements.py
```diff
@@ -27,81 +28,228 @@ def newton(f,x0,fprime=None,args=(),tol=1.48e-8,maxiter=50):
...
+        ):
+    """
+    returns two particlesets, which contain the primaries and the secondaries
+    in binary pairs.
+    """
+    number_of_primaries = len(primary_mass)
+    number_of_secondaries = len(secondary_mass)
+
+    # mass arrays need to be the same length
+    if number_of_secondaries != number_of_primaries:
```
ipelupessy: maybe add another acceptable case: number_of_primaries==1 (or input not a vector) and then generate one set assuming independent (zeroish mass) orbiters

rieder:
One of the tests (test10) is now failing, because it is unitless. Actually, I don't want that one to be successful - AMUSE should never be unitless. Is it OK if I remove that test?

ipelupessy:
you can block it out (don't remove)...I generally try to add tests to make sure code is unit and numpy-array agnostic..

rieder:
Had to transpose a few arrays to get the tests working. 
They are now all OK, except where the tests expect unitless angles (tests 10 (E), 4 and 5 (F)). 
Test 10 results in an error since it tries to assign a unitless value to what should be a mass. I think this should result in an error, and the test needs rewriting.
Tests 4 and 5 fail, since the script now returns angles with unit rad even if the input angles were unitless (and therefore assumed to be in deg, per the previous expected behaviour). So the values aren't equal. If this change can be made to the test somehow, it will pass. Or should the resulting angles be given in (unitless) deg in case that was the input? That would make things complicated, so it wouldn't be my choice...

I think the code is now fine by the way, so it can be merged if it checks out OK for others as well and the tests are fixed.

ipelupessy:
ok, great; can you add the tests file to the pull (with the tests that fail renamed (test10 -> xtest10 or something)?

rieder:
Will do. And I'll add (a) test(s) for the new stuff I added.

rieder:
I think it's ok to merge this now.

ipelupessy:
ok great! I think there is one issue left, since Lucie and other people are using the routines with deg assumed, is that the output orbital elements always have units? I think the it would be good for them  to have the old routine name keep their behaviour (with a deprecation warning). I would name the new one just orbital_elements (ie this contains all the new code, orbital_elements_from_binary call orbital_elements and strips/ converts the angle units)

maybe also ask lucie for comment?

btw: why does orbital_elements_from_binaries not call orbital_elements_for_rel_posvel_arrays? (is there a difference)?

rieder:
Probably good to rename, yes. 
The old functions return values with units now, but default to degrees for output and unitless input. So the behaviour is nearly the same - degrees are a pseudo unit... new functions just use units, without a specific default. 

Adding a deprecation warning would be good. 

Yes, the new function should be used internally - forgot to clean that up. The code is the same...

I'd welcome comments from @jilkoval!

jilkoval:
I haven't been following all the detailed changes here, but If it's not a big trouble to keep the `orbital_elements_for_rel_posvel_arrays` working with the original input, that would be great. I'm not working with it at the moment (at least I think so ;) ) and am happy to adjust it in the future scripts, but some old scripts might break otherwise. I think keeping it working with a warning, as @ipelupessy suggests, is a good solution. Thanks!

rieder:
Indeed. Got rid of duplicate code, fixed an error in the test, now it succeeds.

ipelupessy:
I have made the "old" functions return unitless floats again, I think new_binary_from_orbital_elements
should maybe just not work if the input has units (itsdangerous now, becuase it will work with input as e.g. 1|units.rad, but fail/ give unexpected results if that angle is reduced to unitless (still representing  radians, remember that angles can be converted silently to floats..)

maybe rethink the new function names a bit more (shorter, maybe just generate_binaries??

ipelupessy:
to be clear, for me its ready to merge once we are sure the function names stay the same  (the other thing can be fixed later)

src/amuse/ext/orbital_elements.py
```diff
@@ -10,10 +28,9 @@
 
 from amuse.units.quantities import to_quantity, VectorQuantity
 
-
 def newton(f, x0, fprime=None, args=(), tol=1.48e-8, maxiter=50):
     if fprime is None:
-        print("provide fprime")
+        warnings.warn("provide fprime")
```
ipelupessy: not sure why this is not an exception..

rieder:
I'll try to think of good function names. I'm wondering if there is some convention for function names like this in Amuse - if not, maybe we should think about that?

src/amuse/ext/orbital_elements.py
rieder: Not sure either, seems better to have it as a warning indeed. Anyway, you wrote it ;).

ipelupessy:
for me its ok now...

rieder:
Great! I'm fine with it as well.

PR #138, opened 2017-06-07
Vectorised function to generate binaries from orbital elements

rieder:


rieder:
Note that this is not a merge into the master branch... It actually merges into the other patch branch. Somehow.

PR #141, opened 2017-06-08
Pre-requisites installation for Ubuntu updated

sr6033:
The pre-requisites installation including mpich2 has been updated as per the libraries available in official sites.

### DIFF
--- a/doc/install/install-prerequisites-ubuntu.txt
+++ b/doc/install/install-prerequisites-ubuntu.txt
@@ -27,14 +27,83 @@ For openmpi do::
 For mpich2 do::
 	
 	> sudo apt-get install build-essential gfortran python-dev \
-	  mpich2 libmpich2-dev \
-	  libgsl0-dev cmake libfftw3-3 libfftw3-dev \
-	  libgmp3-dev libmpfr4 libmpfr-dev \
-	  libhdf5-serial-dev hdf5-tools \
-	  python-nose python-numpy python-setuptools python-docutils \
-	  python-h5py python-setuptools
+  	  libmpich-dev \
+  	  libgsl0-dev cmake libfftw3-3 libfftw3-dev \
+  	  libgmp3-dev libmpfr4 libmpfr-dev \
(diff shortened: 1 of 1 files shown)


ipelupessy:
thanks! maybe an ubuntu  version number should be added for the new intstructions (I think the older instructions would work for anything pre 2016 or so?). furthermore: do you know why its necessary now to install mpich2 by hand?? (and do you know if the openmpi instructions still work?)

sr6033:
The installation instruction that I have added will work for any new versions of **Ubuntu(15 or above)**. I cannot say about the working of older instructions for previous Ubuntu versions because I haven't tested those on older versions. 

**mpich2** is not available in the official Ubuntu repos list. I faced this issue while trying to install it via apt-get. That is why it has to be installed manually now.

There is another issue that newer versions of mpich2(specifically 1.3 & above) doesn't have **mpd** anymore. That is the reason I have recommended to download mpich2 ver 1.2 which is an older version. If your instructions of **Amuse** can be changed from **mpd** then new versions will work.

arjenve:
Yes, the latest version of mpich in UBUNTU should be fine  (mpich 3.x is known to work).

We do not depend on mpd,  starting scripts without mpd (hydra is now the default) requires the mpiexec command:

mpiexec python <script.py>

So yes, mpich should be used as is from the ubuntu repository...

ipelupessy:
does the package name "mpich" in that case work?

sr6033:
If the official **mpich** is working then the instructions here: http://www.amusecode.org/doc/install/howto-install-AMUSE.html
 need to be updated as per Hydra.

ipelupessy:
thanks, that page should indeed be updated since mpd is now no longer supported I think (though there was info on the hydra manager, but this should be mentioned first as the default)..I think the prerequisite page for ubuntu still needs some updates..didn't have a chance to test it yet...

ipelupessy:
commit 7a576a2 fixes this now

PR #144, opened 2017-06-16
added license and notice

ipelupessy:
This is a proposal for the license.

I am not sure about the copyright holder.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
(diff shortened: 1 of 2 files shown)


rieder:
APL seems to be a decent choice.
It's a bit tricky though - in principle need consent from everyone who contributed to AMUSE code to implement license... 
I guess up to now the "implied" license was "completely free" / MIT style. 

One thing to consider: some community codes in AMUSE are GPL codes. I'm not completely sure APL and GPL are compatible licenses to the point that the interfaces to those codes can be APL. They may need to be under a different license. This could be very tricky stuff, and probably isn't relevant in practice, but still may need to be considered...

ipelupessy:
good points; one of the objective of this pull is to get input from everybody involved!
about the GPL: I think indeed we can include the GPL codes, but the interface on the "native" (so fortran/C/ whatever) side needs to be in the same licence (or compatible) since these build the executable which link to the (GPL) code; this is no problem;

rieder:
Yes, I think that's right. We're communicating through MPI with those codes, so essentially it's a different thing. 
But we'll need to be explicit about the licenses of those interfaces.

rieder:
Let me just state here that I'm fine with using the Apache license v2.0 where it comes to my contributions.

rieder:
Maybe now (before the book is released) is a good time to add the license?

rieder:
This is probably something we should discuss at the upcoming AMUSE meeting...

stevemcmillan:
Sorry, but I don’t see the subject.  What is the issue here?


Stephen L. W. McMillan
Department Head and Professor of Physics, Drexel University
+1 215-895-2709/2723, FAX +1 215-895-2940, steve@physics.drexel.edu

rieder:
Defining a license for AMUSE. Inti suggested we use the Apache Public License. It would be good to have a decision on this.

spzwart:
Hi Steven,

Sounds like something to discuss in April..

Cheers,

Simon

arjenve:
Yes, we will use MIT or BSD license! this will work with the codes too
as these can have their own license

Arjen

ipelupessy:
@arjenve: shall we merge this or do you want to change the licence (Apache v2, almost the same as MIT)

rieder:
Please make sure to note where code may fall under another license (i.e. included community codes)...

ipelupessy:
@rieder : that is in the NOTICE - can you check if you agree?

rieder:
Looks fine to me, though maybe it would be good to add an exception clause in case there is no explicit license for a community code ('all rights reserved' or something?).

ipelupessy:
@arjenve I am inclined to merge now...

PR #167, opened 2017-10-04
Patch for adding Bonsai2

rieder:
This is a pull request tracker for adding Bonsai2 to AMUSE. 
**Please do not merge yet!**

Currently, this will add Bonsai2 next to the 'old' Bonsai, since the two don't (yet) share the same feature-set. 
Bonsai2 adds support for multi-GPU runs, while stopping conditions are not currently supported.
In a future version, Bonsai2 will also act as an SPH code.


rieder:
To get Bonsai2 working:

```bash
    cd src/amuse/community/bonsai2
    ./getBonsai.sh 
    cd ../../../../
    make bonsai2.code
    # Bonsai requires a lot of stack memory, currently...
    ulimit -s unlimited 
    # Test if it works
    nosetests -sv --nocapture test/codes_tests/test_bonsai2.py
```

This should be integrated into the makefile / interface before merging.

PR #176, opened 2017-11-07
fixes to allow for zero dimensional grids

ipelupessy:
something like: g=Grid(); g.mass=5 (and the equiv. for codes also) can work with this
It is mainly convenience. An alternative is to construct a higher dimensional,
but sized 1 grid)


ipelupessy:
I think it works ok now, but maybe someone should check whether this doesn't break anything or has a performance impact, @arjenve ??

PR #183, opened 2018-01-19
Updated macOS install instructions for macports.

rieder:
Removed many details that are no longer necessary, or outdated.
Updated to most recent gcc version

### DIFF
--- a/doc/install/install-prerequisites-osx.txt
+++ b/doc/install/install-prerequisites-osx.txt
@@ -1,7 +1,7 @@
-Installing on OS X
+Installing on macOS
 ******************
 
-Installing on MAC OS X with MacPorts 
+Installing on macOS with MacPorts 
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 In this section we asume a clean MacPorts installation. The MacPorts build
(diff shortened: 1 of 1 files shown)


arjenve:
Steven,  Thank you for this update!! 

As for gcc7, I do not want to put this as the standard option for AMUSE, it has multiple issues (some optimizations are dubious and C++ does thing subtly different). 
So could you please keep the variants to gcc49 and refer to gcc7 / gcc6 where the options are discussed?

rieder:
Ok, I’ll make that change. 

By the way, I was working on installing Amuse with Python3 today, I made a page for that as well on my Amuse github wiki (https://github.com/rieder/amuse/wiki/Installation-on-macOS-with-Python3). I use virtualenv for this install, it seems to work quite well (some issues with nosetests).

rieder:
Ok, I've changed the default back to gcc 4.9. 
As a sidenote, recent versions of Ubuntu and Debian no longer have packages for gcc versions lower than 6, so we should make sure these recent GCC versions work correctly with Amuse.

rieder:
@arjenve OK to merge?

arjenve:
Yes, Will do!

PR #197, opened 2018-03-18
MOBSE & MOSSE: pull request

GiacobboNicola:


arjenve:
Thank you for the pull request,  A few small comments:

* Some binaries are in the pull request, could you do a 'make clean'  / remove the binaries? (mobse_worker, mobse_worker.dSym directory, src/mobse, src/mobse.dSym)
* Some temporary (editor?) hidden files are also in there for example src/_.star.f.
* Some mac files '.DS_Store'

GiacobboNicola:
Dear Arijen, 

Sorry I have not noticed them.

I do it soon.

Nicola Giacobbo

rieder:
@GiacobboNicola I've submitted a pull request to your branch to include code tests for SSE and BSE. Could you review these before we merge this?

GiacobboNicola:
@rieder Of course I can. I'll do as soon as possible.

rieder:
Thanks. The SSE/BSE tests currently fail in a few cases when I adapt them for MOSSE/MOBSE, in most cases because the result differs from what SSE/BSE would give (which is to be expected), but in a few cases it is because of some other error (probably due to interface differences between MOSSE/SSE and MOBSE/BSE).

GiacobboNicola:
You can find a test file (for MOBSE) in its own community directory. It should work. Let me know if you have problem even with that.

rieder:
Looks fine now. Merging.

PR #200, opened 2018-04-13
fix for python install

ipelupessy:
it seems necessary to specify libdir otherwise basic imports fail (locations in lib vs lib64)
(maybe test this, because it should not be necessary..)

### DIFF
--- a/doc/install/install-python.sh
+++ b/doc/install/install-python.sh
@@ -7,7 +7,7 @@
 #
 
 #APPVER=2.5.4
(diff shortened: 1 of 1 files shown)


arjenve:
Could you please keep the "ibiblio" link for ssl download? (in the past the openssl.org download links turned out to be unstable in that these were arbitrarily moved)

ipelupessy:
I keep the new link for the moment( I am optimistic it will stay around at least as long as the other;-)

Issue #202, opened 2018-04-16
numpy dtype change


rieder:
Numpy issues some warnings of the following type recently: 
```
(...)/src/amuse/datamodel/particles.py:302: FutureWarning: Conversion of the second 
argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will 
be treated as `np.float64 == np.dtype(float).type`
```

We will need to update this to the correct new way of doing things in numpy.

rieder:
h5py fix for the same issue is described here: https://github.com/h5py/h5py/pull/926/files

PR #207, opened 2018-04-23
replaces float with np.floating

rieder:
possible fix for #202 

Not completely sure if this is the right method, but this seems to work (and gets rid of the warning message) after initial testing.

### DIFF
--- a/src/amuse/datamodel/particles.py
+++ b/src/amuse/datamodel/particles.py
@@ -283,7 +283,7 @@ class AbstractParticleSet(AbstractSet):
                     values_to_show.append(format_str11('...'))
                     values_to_show.extend(map(format_str11,quantity[-split_at:].to_print_list()))
                 elif hasattr(quantity, 'dtype'):
-                    if numpy.issubdtype(quantity.dtype, float):
+                    if numpy.issubdtype(quantity.dtype, np.floating):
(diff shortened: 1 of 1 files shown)


arjenve:
please change np.floating to numpy.float 
1) we use "numpy" not "np'
2) " FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`." --> check against numpy.float

rieder:
Oops, I had changed that locally & tested with that but not committed/pushed it to the correct branch...

arjenve:
Still a problem with "floating", that name will be deprecated, the code needs to be "float".

rieder:
so it should be `numpy.float64`?

ipelupessy:
I was under the impression that the proper fix is: float -> numpy.dtype(float).type ??

arjenve:
It is, "numpy.float", I tested this.

So not, "floating" and not "float64", the last one would exclude 32bit floats

arjenve:
Looks good!

ipelupessy:
if its good, lets merge..

arjenve:
Ok, let it be "floating".  Apparently float != floating.

ipelupessy:
can you fix this again in the rieder-patch-float branch?

rieder:
Done

PR #210, opened 2018-04-25
Remove old, unused directory

rieder:


arjenve:
This code is for backward compatibility, we did have a simplex2_5 module for more than 1 year and scripts (not in the amuse sourcecode) may still use it. So, please do not remove this.

rieder:
I'm wondering if there would be a less visible way to solve this, but I guess that will have to wait until the MUSE/AMUSE decoupling.

PR #211, opened 2018-05-01
Update Ubuntu installation doc

rieder:
- Package names have changed very slightly recently, but enough to let the old instructions fail
- Remove reference to extremely old Ubuntu 9.10, no one should use this at this point
- Remove reference to testing being done with MPICH preferentially

### DIFF
--- a/doc/install/install-prerequisites-ubuntu.txt
+++ b/doc/install/install-prerequisites-ubuntu.txt
@@ -1,23 +1,22 @@
 Installing on Ubuntu
 ====================
 
-Installing on Ubuntu version > 10.10
+Installing on Ubuntu version 18.04
 -------------------------------------
 
-In this section we asume a default Ubuntu desktop installation.
+In this section we assume a default Ubuntu desktop installation.
 
 All
 ---
 The prerequisites can be installed with a couple of commands
-on Ubuntu. The only choice to make is between openmpi and mpich2. Most
-of our testing is done with MPICH2 but openmpi should also work.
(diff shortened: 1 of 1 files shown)


rieder:
Note: /usr/bin/easy_install is missing in Ubuntu 18.04.
Maybe we should use `sudo python -m easy_install mpi4py` instead of `sudo easy_install mpi4py`?
This works for me.

ipelupessy:
I am ok with this, maybe add a note for older versions along the lines:
AMUSE can also be installed on older versions (up to 10.04), although in that case the package names may be slightly different.

rieder:
I'll add something along those lines.

Note though that older versions of Ubuntu are not supported by Ubuntu after a very short time (normal releases) or a few years (LTS releases).
This means that installing prerequisites on those releases is a much greater effort, as the normal channels will no longer be available.
So I would suggest we do the same and not support old releases except those still supported by Ubuntu.

arjenve:
I have an old version of ubuntu, and all seems fine here. So I would be happier if we still have the copy+paste line for 10 < ubuntu < 18. (So for example, when we needed to did this the first time we made the 9.04 paragraph)

rieder:
Not sure which old version you use, but in my experience it requires a lot of work to keep an unsupported version working - at least changing the apt sources etc.
Anyway, I have no problem keeping old instructions around other than that displaying everything makes things less clear.
This is something we should address, but this is probably not the right pull request for that.
I'll re-add the 'old versions' section for this one.

ipelupessy:
see #448

PR #212, opened 2018-05-01
Update macOS install doc

rieder:
Change gcc4.9 -> gcc5, since older versions are no longer supported on recent macOS versions.


rieder:
It seems the documentation on the website is still using the older text?

ipelupessy:
I think its updated now?

rieder:
Yes, thanks.

PR #214, opened 2018-05-01
add option to ignore whether worker is present

ipelupessy:
this is a harmless hack to allow the startup of a code from a containerized AMUSE install (from outside this container), for example :
```
s=SSE(channel_type="sockets",debugger="custom",
      custom_exe="singularity", 
      custom_args="exec ./ipelupessy-amuse-singularity-master-42.3.simg",
      worker_code_directory="/usr/lib/python2.7/site-packages/amuse/_workers/",
      must_check_if_worker_is_up_to_date=False,
      try_to_locate_worker=False )
```
the worker_code_directory specifies the location within the container, the two other options are necessary to skip check, since we do not look inside of the container atm
(such checks could also be added ofcourse)

### DIFF
--- a/src/amuse/rfi/channel.py
+++ b/src/amuse/rfi/channel.py
@@ -1061,10 +1061,12 @@ class AbstractMessageChannel(OptionalAttributes):
     def custom_args(self):
         return '--hold -e gdb --args'
     
-    
-        
     @option(type='boolean', sections=("channel",))
     def must_check_if_worker_is_up_to_date(self):
+        return True
+
+    @option(type='boolean', sections=("channel",))
+    def try_to_locate_worker(self):
         return True
     
     @option(type="int", sections=("channel",))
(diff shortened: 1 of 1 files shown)


arjenve:
Wow, this has become a complex function over the years.

Could you maybe make it a special case, as the second and third `if not self.try_to_locate_worker:` statements do not make much sense, the third case will never be hit and the second is arbitrary.

So something like

`if not self.try_to_locate_worker: 
     if len(self.worker_code_directory) > 0 and os.path.exists(self.worker_code_directory):
            full_name_of_the_worker = os.path.join(self.worker_code_directory, exe_name)
            full_name_of_the_worker = os.path.normpath(os.path.abspath(full_name_of_the_worker))
            return full_name_of_the_worker 
     else:
            raise Exception('Must provide a worker_code_directory')
`

Some doubling of code, but the intention is clearer. Also maybe not the name `try_to_locate_worker` but I have no better suggestion yet.

ipelupessy:
no comment on the code complexity ;-) 

in any case, chaned it a bit, and added also the option of skipping the check when providing the full path name..

PR #219, opened 2018-05-19
convert warnings for parameter errors to exceptions

ipelupessy:
I think these warnings are not ok:
- users don't pick up the warnings
- this leada to silent errors
- and unexpected behaviour / hard to debug errors
- ignoring non presence of a certain parameter is rather simple and should be done, in those circumstances that this is ok, by the user


ipelupessy:
I am seriously considering merging this; it keeps on being a serious source of errors in production scripts..

rieder:
I'm OK with that

PR #221, opened 2018-05-29
make parameters updateable, ie make it possible to add a parameter to…

ipelupessy:
… a set after instantiation


ipelupessy:
its an old branch I had - it adds the possiblility to dynamically add parameters, after code has started. At the time I think i chose a different way to solve the problem, and was not necessary in the end.

it may need a bit more testing...

ipelupessy:
this will be rewritten..

PR #222, opened 2018-05-29
Fix flat of quantity

ipelupessy:
this patch makes sure the flat of a quantity is itself also a quantity (or at least more quantity like)....


ipelupessy:
maybe the flatiter_wrapper class definition should also be moved outside the method...

PR #223, opened 2018-06-07
Create CODE_OF_CONDUCT.md

rieder:
Add code of conduct, per Github's recommended community standards (https://github.com/amusecode/amuse/community).
I selected the default contributor covenant, which seems fine for me.
Before adopting, needs a 'proper' email address to contact AMUSE core developer group about misbehavior (I think this should not be the amusecode google group, nor a single developer).

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,46 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as contributors and maintainers pledge to making participation in our project and our community a harassment-free experience for everyone, regardless of age, body size, disability, ethnicity, gender identity and expression, level of experience, nationality, personal appearance, race, religion, or sexual identity and orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment include:
+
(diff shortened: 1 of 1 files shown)


ipelupessy:
to be honest, I don't like these codes of conduct...

rieder:
May I ask why not? :)
My feeling is that it doesn't hurt to have a code-of-conduct (especially if it's a widely accepted one), as in practice you probably won't need to use it (people usually behave themselves), but if they don't it's good to have something to point at.
These kind of documents aren't created for times when things go smoothly, but for when you need some ground-rules to fall back onto.

ipelupessy:
well, if you really want to know;-)...its a bit strange that people suddenly need some outside authority to point at..other than that I find the language in these CODs condescending; they are phrased to give some people a warm and fuzzy feeling of virtuousness, while ignoring nuances like: people should be free to speak their mind and have open communication, even if it means that sometimes feelings get hurt...other than that they are also a bit dishonest, because in the end it boils down to a simple "project management decides"

arjenve:
I concur with Inti, *no* code of conduct.

rieder:
Ok.
I still think it doesn't hurt to have one, but I'll close this PR (perhaps until some day AMUSE has many more developers).

PR #224, opened 2018-06-07
Add issue templates

rieder:
Adds templates for creating issues (per the github recommended community standards):
- bug report
- feature request
- other

This may help create clearer issues, which could reduce the time needed to close them.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,24 @@
+---
+name: Bug report
+about: Create a report to help us improve
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+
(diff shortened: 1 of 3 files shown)


ipelupessy:
personally I don't like it when projects do this (as a user submitting I mean). We are I think not at the stage that we need this...(the only reason to do this is when there are too many poorly phrased issue reports)...

rieder:
I think it's also a way of encouraging issue reporting. It shows that we're actively telling people to tell us when they found a bug, want a new feature, or something else. I think too often people find something but don't bother to report it, and this might just help them to do just that.

rieder:
Also, I really want people to report problems they're having (e.g. with installing / using AMUSE) here, instead of using a mailing list / private emails, as happens more frequently. It encourages better documentation / faster fixing of bugs / more satisfaction from fixing a bug. Maybe psychological effects, but it does work...

ipelupessy:
ok, lets merge this, see how it works!

PR #225, opened 2018-06-07
Create CONTRIBUTING.md

rieder:
Adds contributing guidelines, per the github recommended community standards.
This guideline is modified from the Astropy developer guidelines, reflecting (mostly) current practice for AMUSE.

We should discuss if these guidelines reflect what we want before merging, and specifically we should look at documentation requirements. Possibly, we could just completely adopt the Astropy developer guidelines, although those are currently much more strict than ours...


rieder:
Thanks. I figured specifying that label might be helpful in encouraging such kind of commits (we could use them, to be honest), but I don't mind leaving it out.

Issue #236, opened 2018-06-19
getitem for vector quantity fails when dtype is object


ipelupessy:
**Describe the bug**
getitem (ie subscripting) fails for vector quantity with a object data type

**To Reproduce**
```
>>> a=numpy.arange(10, dtype=object) | units.m
>>> print a
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9] m
>>> print a[1:4]
[1, 2, 3] m
>>> print a[2]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/inti/code/amuse/amuse/src/amuse/units/quantities.py", line 590, in __getitem__
    if number.shape:
AttributeError: 'int' object has no attribute 'shape'
```
**Expected behavior**
should not fail, just return the requested object, this will/may be important when fixing #155

ipelupessy:
one of the ways to fix #155 involves unit_with_specific_dtype whihc owuld trigger this...

ipelupessy:
automagically fixed (at least the example given)...

PR #238, opened 2018-06-20
simplify getitem, fixing shape error in case of object dtype

ipelupessy:
might have slight performance penalty


ipelupessy:
should fix #236

ipelupessy:
ok, this  reverses a change made in 185153f775bc1cb229351047a3b4b84a78b03232, which consists of a bunch of performance related fixes...

ipelupessy:
@arjenve: the problem was in the if number.shape - number would not be a numpy object if array was of object type..I don't see another way to fix this than using new_quantity (well copying its code verbatim); is this ok (with the call)?

Issue #186, opened 2018-01-25
Building a single code with python3


rieder:
With python2, it's possible to build only a single code, or even just the Amuse framework:
```make framework``` or e.g. ```make ph4.code```

Since the installation for python3 is quite different, I'm wondering how to do something similar there. Building all the codes can take a long time...

ipelupessy:
I agree, for python 3 it would be much preferable to be able to use the same workflow....

ipelupessy:
half done, remains: #282

PR #256, opened 2018-06-26
Single build for Python 2 and 3

evertrol:
(*do **not** merge; preview*)

This is a preview for a PR for a single build system that works for both Python 2 and Python 3.
This removes the `support3/` directory, and makes the modules in `support/` Python 3, with Python 2 backward compatibility.
Build files such as `setup.py` and `Makefile` also remove their dependencies on `support3`.

At the moment, many parts are not built (i.e., less parts are built than with the `support3` package), for both Python 2 and 3. There'll be parts I've overlooked in the conversion, but I've put the PR here so people can have a peek.

At the moment, during a build, the `2to3` script is run to convert all the Amuse code to Python 3 (so this can be built in-place). This change can afterwards be easily undone with `git checkout -- src`.


ipelupessy:
I think in case of python 3 you don't want in place build...only build and install

evertrol:
This now provides single build mechanisms for Python 2 and 3. For example, run configure with the relevant Python executable set:
```
PYTHON=/usr/local/bin/python3.6 ./configure
```
and each of the following works:
```
make build
make framework
make <something>.code
```

At the moment, all Python 2 builds are in-place (as was the default), while all Python 3 builds are out-of-place (to not mess up the git source tree). See also #271 .

The work-around for building a single code is similar as for `setup.py build_codes`, since the relevant Python files have to be copied first to a build dir, before converted to Python 3. I've tried to copy only the relevant files, and avoid copying and converting them again if another single code is build. But "relevant files" may be too few, so perhaps all files (`*.py` and a few `*.template` files) should be recursively copied (the `codes/` subdirectory could be excluded, but not so e.g., the `community/` subdir).
See also the commit message for fc6682d6890f012677f5bd11ce7ba9aaf590ca7f

evertrol:
Might be good to move the last commit (fc6682d6890f012677f5bd11ce7ba9aaf590ca7f) into a separate PR. That new PR can then also be used to address #186 .

evertrol:
Replaced by #273 .

PR #261, opened 2018-06-26
Switch to container-based run, which boots faster

egpbos:
Replace sudo apt-get calls with the apt addon, which doesn't need sudo anymore, making the builds faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,24 @@ python:
 
 os:
   - linux
+#  - osx
 
-dist: trusty
-
-sudo: enabled
+addons:
+  apt:
+    packages:
+    - build-essential
+    - python-dev
+    - gfortran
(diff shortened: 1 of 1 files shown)


ipelupessy:
for me it looks ok..(but not much experience with travis)

egpbos:
Most important thing is that it works ;) It boots up a bit quicker, see https://docs.travis-ci.com/user/reference/overview/#Virtualization-environments

ipelupessy:
ok, if it works its ok, it has a little less resources though...

egpbos:
Yes, we should definitely keep that in mind for when we want to do more demanding tests. It is also possible to only run on the higher resource (sudo-enabled) machines for those specific cases using a build matrix. I have started a build matrix in #262 if you want to see an example of this. Off topic for this PR though.

Issue #247, opened 2018-06-25
Add community code tests to Travis


egpbos:
Some of these tests take very long, so maybe not test everything on all platforms. Alternatively, we could build a test matrix with separate builds for separate (groups of) tests.

The CUDA codes will not work at all on Travis, so skip those.

arjenve:
I do not understand the "contributor code"  part...

egpbos:
Sorry, I meant community code, changed the title.

Issue #235, opened 2018-06-16
Installing AMUSE on a Mac where Anaconda Python Distribution Installed


OrbitalMechanic:
I'm interested in installing AMUSE on a MacPro (Late 2013) that run MacOS X Sierra (ver. 10.12.6) that has Python 3.6.5 Anaconda distribution installed. Is there a procedure for installing AMUSE under Mac OS X with the Anaconda distribution of Python already installed that does not require the use of MacPorts?

Please advise.

ipelupessy:
Myself, I am not too familier with mac installs (or anaconda)..
@rieder: have you tried this?

rieder:
I’m in the process of trying :). 
I’ll hopefully have something more to say later this week.

rieder:
I tried this by installing the prerequisites through conda.
Unfortunately, all the Fortran-based community codes fail to build. This is because the openmpi I installed (from conda-forge) doesn't have fortran support - probably because it's built for the Xcode compilers which lack this (maybe this shouldn't be a blocking issue for installing any part of AMUSE, but that's a separate issue). 
If there is an openmpi (or mpich) with fortran support which can be installed through conda, I guess it could work. I'll keep looking for now, but if anyone knows of such a package, please let us know.

egpbos:
Hm, I deduced from [this](https://github.com/conda-forge/openmpi-feedstock/search?q=fortran&unscoped_q=fortran) that the conda-forge openmpi would be fortran enabled. Indeed the fortran codes build on my machine (High Sierra 10.13.5, Xcode 9.4.1, clean conda env with python2.7). However, I also have Macports gfortran installed, so I'm not 100% sure which fortran gets used.

egpbos:
I also installed gfortran_osx-64, that seems to make sure that indeed the conda fortran compiler is used instead of the macports one.

rieder:
If that works, great!
Did you install gfortran_osx-64 through Anaconda as well?

egpbos:
Ok, just installing pkg-config does not magically fix everything.

I did find a way to add the necessary rpath to a complaining executable after the fact. One example is `hermite0_worker`:

```
cd src/amuse/community/hermite0
install_name_tool -add_rpath ${CONDA_PREFIX}/lib hermite0_worker
```

[...]

PR #262, opened 2018-06-26
macOS test build on Travis

### LINKED ISSUE #245: Add macOS testing to Travis
Following #241 #244

egpbos:
Fixes #245

The setup with `matrix: include:` is necessary because `language: python` builds are not possible in combination with `os: osx`, so we need to do `language: generic` for `os: osx`, instead of just using `language: python` for the entire file.

This also necessitates some reordering of the Linux specific commands to fall under the matrix. When we will add more matrix options (e.g. different versions of Python), this will become a lot of copy-pasting. At that point, it may make more sense to do some Bash scripting in the before_install, install and script steps to detect the OS and other build parameters in order to select different install/run options. For an example setup see https://github.com/egpbos/xtensor-fftw/blob/master/.travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,36 @@
-language: python
-python:
-- '2.7'
+env:
+  global:
+    - MINICONDA_VERSION="latest"
+    - MINICONDA_PYTHON_MAJOR=2
+    - MINICONDA_LINUX="Linux-x86_64"
+    - MINICONDA_OSX="MacOSX-x86_64"
 
-os:
-  - linux
-#  - osx
+matrix:
+  include:
+  - language: python
+    python:
+    - '2.7'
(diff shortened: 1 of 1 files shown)


egpbos:
Does anyone feel like reviewing this? If not, maybe just merge, it seems to work. Note that this does not do the community code builds yet, that is another issue (#247).

The first commit is the same as that in #261 (this is a branch off that branch), so please merge that PR before this one.

egpbos:
This is also the first part of a fix for #235. Based on this example, after including the community code builds, we can write some documentation for this process, which is basically just copy-pasting the commands used in the Travis build(s).

ipelupessy:
looks ok, I think the test10 and 11 are just floating precision issues??

Issue #267, opened 2018-06-27
split framework off into src/muse


ipelupessy:
general outline: move everything framework related to src/muse, replace everything in amuse that is moved with imports from framework src/muse. Make sure autogenerated files are generated in correct location. Some deprecated features, import locations do not need te be copied. Some namespaces/classes will have generic and amuse versions, eg: amuse.units vs muse.units or amuse.datamodel.Particles vs muse.datamodel.Particles.

detailed todo:

- [ ] move and adapt src/amuse/__init__.py, src/amuse/amuserc
- [ ] move src/amuse/rfi
- [ ] move src/amuse/datamodel, trim down data storage classes to generic versions copying amuse specific stuff back to derived classes in src/amuse/datamodel, 
- [ ] move src/amuse/support. Note some older deprecated import wrappers are still in there - these do not be copied
- [ ] move src/amuse/test
- [ ] move src/amuse/units
- [ ] move src/amuse/io
- [ ] check src/amuse/ext - small subset may be moved to muse
- [ ] check community: some support code and/or codes (Hop) may be important for framework
- [ ] check couple (same story)
- [ ] check ic

ipelupessy:
this issue is no longer relevant since packages split. residual task is to make a better seperation between astrophysics and framework code. This has to be done within the amuse namespace since muse is taken. So make a new issue for this...I keep this issue a little while longer because the todo is still a good list of things to keep in mind...

PR #270, opened 2018-06-27
make muse framework directory

ipelupessy:
initial muse directory...not sure yet whether to create seperate pulls for every item in #267

### DIFF
--- a/src/muse/__init__.py
+++ b/src/muse/__init__.py
@@ -1,0 +1,24 @@
+"""
+MUSE framework: The MUlti-physics Software Environment
+
+The aim of MUSE is to provide a software framework in which existing codes 
+can be accessed in a homogeneous way in order to perform state-of-the-art 
+simulations of a wide range of different physical systems. It serves as the 
+base framework for AMUSE, OMUSE etc
+
+It contains several packages, most notably:
+units     - MUSE uses quantities, i.e. a number (or array) with a unit attached, 
+            instead of just numbers (vital when coupling different codes!)
+datamodel - defines particles and grids, on which data (quantities) can be stored
+io        - how to read and write data in several file formats
+
+"""
+import numpy
+
+def numpy_fix():
+    try:
+      numpy.set_printoptions(legacy='1.13')
+    except TypeError:
+      pass
+      
+numpy_fix()


rieder:
Should this (at some point) be in a separate repository?

arjenve:
I would not make it a subdirectory in the amuse code instead a new repository as Steven also pointed out.

egpbos:
I think indeed a new repo is the only way if things like Omuse and Hymuse are to build on Muse as well. Mono-repos do have advantages as well though, for instance with CI: Travis wouldn't trigger an Amuse build anymore when Muse changes things, so you will probably miss more downstream breaks.

ipelupessy:
eventually you would want a new repository; but I think this is a way to do the split quickly (basically everything should keep on working at every step, especially for people currently working with amuse, omuse), plus: I am not sure we have decided how to organize the two packages (options are..? I can think of: 1) muse with amuse, omuse as extensions, 2) seperate packages with muse as prereq.

evertrol:
Would a [submodule](https://git-scm.com/docs/gitsubmodules) be a good way to do this?
A separate repository, but it remains the main code. Parts (files & subdirectories) may be easily moved from the main repository to the new repository at leisure, while nothing on disk changes.
Main thing is that people will need to keep the submodule up to date if they use the source code.

rieder:
Not sure... At some point, I imagine that "muse-core" would be pip-installable, and a prerequisite for 'amuse' (and the others). I guess until that time it's best to keep it as one repository.

egpbos:
I like the concept of git submodules, but I have yet to find a convenient way of working with them. Keeping submodules up to date manually is too easily forgotten.

arjenve:
The submodules will integrate the code management system into the code itself, this is never a good idea. Splitting off a separate project is the end goal so let's just do that. I.E. copy all of amuse and start removing comunity codes etc etc. When that is done, start using that code in amuse.

ipelupessy:
@arjenve: wouldn't that leave you with two code bases and/or multiple branches that need to be synchronized for the duration that:
1) the framework package is developed
2) amuse is stripped of the framework components
??

ipelupessy:
I am experimenting with arjens approach at: https://github.com/ipelupessy/amuse-framework

PR #272, opened 2018-06-27
Dynamic build tests refactoring and fixing

### LINKED ISSUE #264: Fix dynamic build tests by adding LDFLAGS
Make a new submodule amuse.test.compile_tools which includes fix for ld_flags and then manually replace copy-pasted test codes by these tools. Follows up on https://github.com/amusecode/amuse/issues/258#issuecomment-400273654

egpbos:
For #264. Work in progress, but putting it here so people can already have a look.


src/amuse/test/compile_tools.py
```diff
@@ -0,0 +1,455 @@
...
+    if process.returncode == 0:
+        wait_for_file(exename)
+
+    if process.returncode != 0 or not (os.path.exists(exename)
+                                       or os.path.exists(exename+'.exe')):
+        print("Could not compile {0}, error = {1}".format(exename, stderr))
+        raise Exception("Could not build {0}, error = {1}".format(exename,
+                                                                  stderr))
+
+    print(stdout)
```
ipelupessy: prints?

test/codes_tests/test_c_implementation_simplified.py
```diff
@@ -3,6 +3,7 @@
 from amuse.support.interface import InCodeComponentImplementation
 
 from amuse.test.amusetest import TestWithMPI
+import amuse.test.compile_tools as compile_tools
```
ipelupessy: minor: import style

src/amuse/test/compile_tools.py
egpbos: yeah, they were there already

test/codes_tests/test_c_implementation_simplified.py
egpbos: What would you prefer?

src/amuse/test/compile_tools.py
ipelupessy: zap them;-)

test/codes_tests/test_c_implementation_simplified.py
ipelupessy: there is no difference between them? I think they look best when all the same;-)

egpbos:
Apparently, some compile_tests don't work on Travis either, so it's not just my Mac setup that was at fault. Any existing issues on this or should we start a new one?

egpbos:
First implementation of everything done. Now on to bug fixing.

ipelupessy:
fixed two errors; cxx_compile could not find header it just wrote, added extra args for this and stopping condition compilation had undefined rootdir

egpbos:
Removed WIP tag, as we're now in bug fixing phase (as opposed to "doing the basic intended work" phase).

egpbos:
Ok, no more strange hangs in the Travis build. There are still failing tests. The question now is: which of these tests were already failing in the master branch?

egpbos:
I find it quite hard to compare to the buildbot.

In any case, perhaps this can all be merged now anyway, since the main goal of the PR (and #264) was to fix that the builds could not be run on Conda due to missing LDFLAGS. This is now fixed. The fact that some tests still fail I think should be the topic of a new issue/PR. What do you think @ipelupessy?

ipelupessy:
revert because it was broken, but also because better put all build_worker in compile tools..

ipelupessy:
for linux its down to an error for cython builds and 2 java fails, will merge if it passes w/o these

Issue #271, opened 2018-06-27
Make building out-of-place the default


evertrol:
The default for building with Python 2 is in-place. While this is easiest, in particular for building single codes, the default should probably be an out-of-place build.

ipelupessy:
I think you want same behaviour for python2 and 3, and since python 3 does not support in-place build due to 2to3, this is reasonable, but this will probably invite discussion! (which is good)

evertrol:
On the other hand, if a Py2/3 compatible code base could happen relatively soon (at least for relevant files), this issue may become largely moot.

egpbos:
Even with a 2/3 compatible code base, out-of-place builds make management of the repo easier (no long gitignore list, just ignore `build/`; easy restarting from scratch by deleting the build directory; easy multiple builds with different configurations), so I would still vote for it.

rieder:
So would I.
virtualenv makes it easy to install in a separate directory if this is desired anyway.

ipelupessy:
closing this; out of place is the default if using setup.py, other issues are fixed with combinations of pip install [-e] etc

PR #273, opened 2018-06-28
Merge support and support3 directories

evertrol:
This merges the support and support3 directories, and makes the main make targets the same for Python 2 and 3. Python 2 will still build in-place, while Python 3 code is build in a build/ directory (see also #271 ).

This should also make it possibly to add Python 3 to Travis, using the same workflow. I haven't added this yet, but if people agree, I can add that change as an extra commit to this PR.

This PR replaces #256.


ipelupessy:
atm its not correctly converting interface python files to python 3...

ipelupessy:
for me its good to merge..tested with inplace and out of place python2, out-of-place python3

PR #275, opened 2018-06-28
Single code build py23

evertrol:
This is an initial PR to address #186 .
This will need some further work, in particular to address the question what is the favourite workflow here (for Python 3):
- build a code in a separate build directory, and set `PYTHONPATH` to pick up the packages from there.
- install the full framework first, then install single codes as requested. There is no install option yet for a single code, so this will have to be changed. 

The second option has my preference, although this current preview PR works more towards the first option. This is partially because this seems to be main workflow with Python 2, where everything is done in-place, and no installation is used (for neither the framework nor single codes).

NB: this PR relies on #273 .


support/setup_codes.py
```diff
@@ -1207,14 +1236,54 @@ def subdirs_in_lib_dir(self):
...
+        """
+        srcdir = os.path.dirname(self.codes_src_dir)
+        destdir = os.path.dirname(self.codes_dir)
+        if srcdir != destdir:
+            owd = os.getcwd()
+            os.chdir(srcdir)
+            amusebuilddir = os.path.join(destdir, 'amuse')
+            # Only Python 3.5+ supports '**', with recursive=True
+            # (even if this function is likely only run by Python 3.5+)
+            files = glob.glob('*.py')
```
evertrol: Better to use a built-in distutils function that grabs these files from setup.py or MANIFEST.in.
Needs some digging into the distutils code, e.g. the build command.

ipelupessy:
superseded by ed26f4f45e1305261c9bc2902ba631ee7ed8354a

ipelupessy:
only thing left, install of single code

PR #283, opened 2018-07-11
Updated the install scripts for Python2 and Python3. 

ipelupessy:
Furthermore, the script for Python2 can also install Python3 now, so in the future a single script could be used.


doc/install/install-python.sh
```diff
@@ -139,9 +130,8 @@ else
 	    --enable-unicode=ucs4\
 		--program-suffix=.exe ;
 fi
-
 make 
 make install
-make 
-make install
-echo "..Done"
+echo "Done"
```
ipelupessy: I think the make and make install was done twice for a reason, but I cant remember what that was??

Issue #284, opened 2018-07-14
Simplex compilation errors


stevemcmillan:
Simplex fails to bulid using Ubuntu 16 and 18 with apt prerequisites, and Macos High Sierra
with Macports prerequisites.

Saying 'make clean all' in the simplex directory should build the simplex_worker code.  The underlying cause of the problems seems to be that  **configure**  doesn't identify the proper HDF5 paths and flags.

There are multiple sources of error:

1.  On Ubuntu 16 and 18, the Simplex HDF5 write functions fail to compile because apt installs hdf5.h in /usr/include/hdf5/serial which is not picked up by configure.  The correct line in config.mk would be

        HDF5_FLAGS=-I/usr/include/hdf5/serial

but the actual assignment is blank.

This isn’t a problem in MacOS because Macports installs hdf5.h in /opt/local/include, which is automatically detected by the compiler.

2.  Even with the header location fixed, compilation fails on all three OSs because -lhdf5 does not appear anywhere in the Makefile. In addition, in Ubuntu 16 and 18, apt places the library in /usr/lib/x86_64-linux-gnu/hdf5/serial, which again is not picked up by configure.  Macports installs libhdf5.a in /opt/local/lib, which is properly detected by the compiler.  Thus the correct config.mk entries would be

        HDF5_LIBS=-L/usr/lib/x86_64-linux-gnu/hdf5/serial -lhdf5

for Ubuntu 16 and 18 and

        HDF5_LIBS=-L/opt/local/lib -lhdf5

in MacOS (the -L part isn’t actually needed).  Again, configure leaves these assignments blank.

3.  Finally, the Makefile uses make syntax incorrectly.  The line

        HDF5_LIBS ?= something

should set HDF5_LIBS if it isn't already set, providing a means of correcting configure shortcomings.  However, once config.mk is done, it leaves HDF5_LIBS set equal to the null string, so the  ?=  line does nothing.

arjenve:
Thank you for the extensive report!

Could you please try to run configure with:

./configure --with-hdf5=/usr/lib/x86_64-linux-gnu/hdf5/serial

Although it probably won't work (very strange practice to put libraries in directories like this)

In Mac Os this should work

./configure --with-hdf5=/opt/local

(if the headers are in /opt/local/include and libs in /opt/local/lib)

I'll see if there is any other way to do this optional setting in the Makefile

rieder:
Related to #209?

PR #286, opened 2018-07-19
change ?= to an ifeq in the Makefiles for issue #284

arjenve:
Convert all ?= lines in the Makefiles of the codes (which will not work if the variable is set but empty) into 3 lines with ifeq.

### DIFF
--- a/src/amuse/community/phiGRAPE/Makefile
+++ b/src/amuse/community/phiGRAPE/Makefile
@@ -1,16 +1,28 @@
 # standard amuse configuration include
 # config.mk will be made after ./configure has run
-AMUSE_DIR?=../../../..
+ifeq ($(strip $(AMUSE_DIR)),)
+    AMUSE_DIR = ../../../..
+endif
 -include ${AMUSE_DIR}/config.mk
 
 
-MPICXX ?= mpicxx
-MPICC ?= mpicc
-MPIFC ?= mpif90
-MPIF90 ?= $(MPIFC)
+ifeq ($(strip $(MPICXX)),)
+    MPICXX = mpicxx
+endif
+ifeq ($(strip $(MPICC)),)
+    MPICC = mpicc
+endif
+ifeq ($(strip $(MPIFC)),)
+    MPIFC = mpif90
+endif
+ifeq ($(strip $(MPIF90)),)
+    MPIF90 = $(MPIFC)
+endif
 FC = $(MPIF90)
 CC = $(MPICC)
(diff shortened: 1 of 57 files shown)


ipelupessy:
I think this is overkill (makes a lot of Makefiles ugly to read)..better to fix few cases where it matters (if it ever does, simplex case I don't think qualifies because the root cause was the config failing (which could be set by hand))

in any case: there are errors in the conversion...

src/amuse/community/phiGRAPE/Makefile
```diff
@@ -24,35 +36,58 @@ CODELIB=$(BUILDDIR)/libphigrape.a
...
 
-SAPPORO_LIBDIRS ?= -L$(AMUSE_DIR)/lib/sapporo_light
-OPENMP_CFLAGS ?= 
-SAPPORO_LIBS ?= $(SAPPORO_LIBDIRS) -lsapporo $(OPENMP_CFLAGS)
-GRAPE6_LIBDIRS ?= -L$(GRAPE6)/lib
+ifeq ($(strip $(SAPPORO_LIBDIRS)),)
+    SAPPORO_LIBDIRS = -L$(AMUSE_DIR)/lib/sapporo_light
+endif
+ifeq ($(strip $(OPENMP_CFLAGS)),)
+    OPENMP_CFLAGS = SAPPORO_LIBS ?= $(SAPPORO_LIBDIRS) -lsapporo $(OPENMP_CFLAGS)
```
ipelupessy: this doesn't look right...

ipelupessy:
I think the better alternative is to not export unset variables in config.mk, or to check whether variables that should be set, are....

(really, the resulting makefiles are clearly not written by/for humans..)

arjenve:
Well yes, the configure could not find the libs. And if you then change the Makefile in testing you get very hard to understand and debug behavior. (I.E. your change will not be used for an unknown reason)

src/amuse/community/phiGRAPE/Makefile
arjenve: Yes I see, clearly not right! (also original had
 OPENMP_CFLAGS ?=
So optional not setting, which is very strange)

I'll update this one

ipelupessy:
also I am not sure if the original behaviour is in many cases not the exact desired behaviour, e.g. some flag being empty can be the desired contents of the flag...

arjenve:
Well yes, that's also true. Lets bin this pull request.

ipelupessy:
ok, lets do that;-)

PR #296, opened 2018-09-05
improve speed when sending large number of strings over mpi channel

ipelupessy:


ipelupessy:
it would be nice if the sockets channel can also be sped up for string transfers (I think there was already an issue for this..)

arjenve:
Looks good and good idea!

PR #298, opened 2018-09-06
fix to make receive strings faster

ipelupessy:


ipelupessy:
this fix assumes that the strings themselves do not contain \x00 - this is mostly ok but e.g. for pickled objects with binary protocol this is not true...

ipelupessy:
no merge for above reason..

PR #300, opened 2018-09-13
Socket string fix

ipelupessy:
fix memory leak in sockets string code
make socket string message similar to mpi
various optimizations


ipelupessy:
the brokenness is on older compilers (newer work)

ipelupessy:
think its ok now

PR #301, opened 2018-09-13
Socket channel boolean fix

ipelupessy:


ipelupessy:
this one should also be ok

PR #308, opened 2018-09-24
Multiple parameter sets

ipelupessy:
this patchset allows the definition of multiple parameter sets on a code as a means to organize large parameter sets.

The rationale is that this is simialr for compound codes when e.g.
code.gas_parameters -> referal to code.gas_code.parameters
code.grav_parameters -> etc

defines parameter_set_names() function 

I am not sure yet if this is worth the complication


arjenve:
Yes, I do not like the second one  (all codes should at least define a 'parameters' attribute)

ipelupessy:
reverted the second part!

PR #311, opened 2018-09-27
added citation metadata

jspaaks:
Hi @ipelupessy,

this PR adds a CITATION.cff metadata file for citation. I then used ``cffconvert`` from PyPI to generate a ``.zenodo.json`` from it, such that we can tell Zenodo about the metadata we want to associate with each release.

The set of authors in CITATION.cff needs checking, in order to determine we included the correct people. I don't know if you want to include all commiters as well, although I feel including them is a token of appreciation for their work. BTW I used the order given here https://github.com/amusecode/amuse/graphs/contributors to determine the order the authors should be in.

Also, please check the orcids and affiliations (i just did a superficial search which hopefully led me to the right people, but it's worth checking.)

Finally, the metadata assumes that there will be another (pre)release ...today. I've labeled it ``11.3`` to be consistent with your naming scheme.

Let me know if I can help with anything,
-Jurriaan

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -53,7 +53,15 @@
         }, 
         {
             "affiliation": "Leiden University", 
+            "name": "van der Helm, Edwin"
+        }, 
+        {
+            "affiliation": "Leiden University", 
             "name": "Beedorf, Jeroen"
+        }, 
+        {
+            "affiliation": "Netherlands eScience Center", 
+            "name": "Bos, Patrick"
         }, 
         {
             "affiliation": "Leiden University", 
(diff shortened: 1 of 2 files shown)


ipelupessy:
ok, thanks! I did notice that the autogenerated authorlist was not good....I think we do want to change the order , my proposal would be:
Portegies Zwart, van Elteren, Pelupessy, McMillan etc
and also a lot of authors of contributing packages are not included (since committed themselves), we are still thinking how to solve this...

rieder:
I think the author list is (mostly) OK for the "core" AMUSE, but indeed not for the current, "full" AMUSE distribution.

ipelupessy:
ok, will also check the old svn committers (e.g. Nathan, Marcell should also be on it)...in any case: what is the convention for the affiliation: is it the current or the one when contributing?

jspaaks:
Good question, i would think it's the one while contributing.

ipelupessy:
the git log gives around 70 committers (with some duplicates)..

jspaaks:
@ipelupessy BTW theres no need to write the ``.zenodo.json`` file by hand, just use ``cffconvert`` from PyPI:
```bash
pip install --user cffconvert
cd thedirthathasacitation.cff
cffconvert -f zenodo -ig > .zenodo.json # but the description section likely needs editing, cffconvert is a bit overenthusiastic when it comes to escaping characters
```

spzwart:
Ok like the idea of a credit section.

there are so many people contributing, even without knowing, but just by 
making their codes available. I think that they deserve some credit too.

ipelupessy:
I have added a few more people; its still a bit arbitrary, but I have added those
committer which have more than 10 commits and/or who have had significant  contribution
when (long time) working with AMUSE

ipelupessy:
this would leave e.g. authors of the community codes (whom may not even be aware of their code being included) for the credits section (this could be tens or hundreds of people)

ipelupessy:
btw, who was svn user "punzo" ??

rieder:
Davide?

PR #323, opened 2018-10-09
add (ro) to read only attributes in grid string

ipelupessy:


ipelupessy:
example print out:
```
CartesianGrid(10 x 20) ( plate_surface__temperature, x (ro), y (ro) )
```

PR #328, opened 2018-10-09
Mher/style fixes examples syllabus

mherkazandjian:
this MR increases the ``pylint`` score from ``-16.38/10`` to ``-10.0/10``


ipelupessy:
syllabus style changes should be coordinated with @spzwart and @stevemcmillan

PR #327, opened 2018-10-09
Mher/style fixes tutorial validation

mherkazandjian:
this MR increases the ``pylint`` score from ``-16.48/10`` to ``4.13/10``

PR #326, opened 2018-10-09
Mher/style fixes 1

mherkazandjian:
Hi,

I did major style fixes for the scripts under examples/application. 
The changes raise the ``pylint`` score from ``-4.01/10`` to ``+4.13/10``.
pylint also complains about some other code quality issues, but pep8 compatibility
is close to ideal now.
  
I also included the encoding (just for completeness) and the __future__ print to
enabled python3 compatibility.

PR #329, opened 2018-10-09
WIP: Mher/style fixes examples simple

mherkazandjian:
this MR increases the ``pylint`` score from ``2.13/10`` to ``6.13/10``

### DIFF
--- a/examples/simple/agb.py
+++ b/examples/simple/agb.py
@@ -1,8 +1,8 @@
+# -*- coding: utf-8 -*-
 """
 Evolves an the sun in it's AGB star phase using SSE
 """
 from __future__ import print_function
-
 import numpy
 from matplotlib import pyplot
 
--- a/examples/simple/agb.py
+++ b/examples/simple/agb.py
@@ -1,4 +1,4 @@
-# -*- coding: utf-8 -*-
+# -*- coding: ascii -*-
 """
 Evolves an the sun in it's AGB star phase using SSE
 """
--- a/examples/simple/binary_population_synthesis.py
+++ b/examples/simple/binary_population_synthesis.py
@@ -1,4 +1,4 @@
-# -*- coding: utf-8 -*-
+# -*- coding: ascii -*-
 """
 Generates a grid of binaries with different, primary mass, mass ratio
 and separation and evolves these over time.
--- a/examples/simple/binary_population_synthesis.py
+++ b/examples/simple/binary_population_synthesis.py
@@ -26,7 +26,7 @@ def multidimensional_meshgrid(*arrays):
     range in one dimension.
     """
     reversed_quantities = tuple(reversed(arrays))
-    lengths = list(map(len, reversed_quantities))
+    lengths = [len(x) for x in reversed_quantities]
     dim = len(reversed_quantities)
 
     size = 1
--- a/examples/simple/binary_population_synthesis.py
+++ b/examples/simple/binary_population_synthesis.py
@@ -205,6 +205,6 @@ if __name__ == "__main__":
 
     print("generated a population of", len(binaries), "binaries")
 
-    evolve_population(binaries, stars,  1 | units.Gyr, 250 | units.Myr)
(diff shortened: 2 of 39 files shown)


mherkazandjian:
i think the utf-8 encoding line is not necessary for python3 compatibility but is need in case a source
file includes a utf-8 character.

yes, i agree that pep8 is not holly and there are situations where the situation would be better off
when the pep8 guide is not followed. i don't think examples in general fit this category. the examples
are the first thing that a user looks at to find out how to do something, as i did, and some of the examples
(not sure who contributed them) looked bad when it comes to the style and there are some examples
that have undefined variables, that i caught because of the style checker.

i recommend including a style checker before each pull request is merged. i.e a pull request that
makes the style score worse, should not be merged until at least the score before the merge is 
maintained (that is how things are done in projects such as scipy and many big projects such as amuse).
it is not a waste of time to pay attention to the style, it is good practice :) 

the changes were too many to comment on them, i'll summarize them in the description box.

ipelupessy:
how configurable would this be (e.g. we need to ignore community code)?

arjenve:
I do not like the extra utf-8 line, we keep all code ASCII and except for one file all character codes < 127.

arjenve:
As for the __future__ imports, please do not add these if the file does not even contain a print functions.

The print function is sort of ok for the simple examples but not in other parts of the code. 

In general we do not want to create code that is *not* python 2 (the future stuff) but also *not* python 3.
I.E. no dependency on python 3 in python 2 libraries (six etc) and no future imports if at all possible.

It seems that in some files the future import for print was added but also extra parenthesis to the print functions (see isochrone.py).

The no "automatic" tools changes has mainly to do with the ability to be able to read / review the changes. (but we have had changes in the past that where large and touched many parts of the code so this rule can be stretched when necessary).

rieder:
@mherkazandjian - thank you for contributing your time to making AMUSE more usable!
As @ipelupessy and @arjenve already mention, there are some points where your fixes may be more pep8-style compliant, but conflict with the (sometimes informal) AMUSE style. 

I'd be happy to help modify this pull request (and #326 #327 #328) so that we hopefully reach the optimal balance between the two :). I'll look over the changes you've submitted and contribute my changes hopefully over the next days or so.

Also, I think it would be a nice idea to integrate a style checker, as a way of increasing awareness - I've done this for myself, and it does make me more aware of where my code needs e.g. more clarity, documentation or even bug fixing.

We should not enable all possible tests though - some go against the code style we like for AMUSE, and others are perhaps a bit too strict (so I agree with @ipelupessy on not quoting pylint score - this is by definition arbitrary). We should find some balance between creating better code and lowering the threshold for people to contribute code. This may be better discussed offline during the next AMUSE meeting/hackathon/sprint.

mherkazandjian:
Hi @rieder 

cool. i enjoyed this exercise and i'll be happy to look at your comments. I completely agree about the 
balance and this should be stated clearly. so i think http://www.amusecode.org/doc/reference/style_guide.html
should be improved too. i.e add more examples of "do"s and "dont"s.
(maybe we can pick examples from the list of my pull requests and maybe also revert some of the changes that i did that look like overkill).

rieder:
Hi,
As you can see from the large amount of commits above, I went through all the changes in the PR, and where necessary I made some changes (removing unnecessarily added brackets in print functions, keeping ascii as the encoding, and keeping python2 functions where problems could arise with the python3 variant). 
I think this can now be merged, but @arjenve could you have a final look as well?
Thanks!

mherkazandjian:
Thnak you @rieder ,
i went over your changes. tnx for catching the extra parenthesis. ok for the rest.
do you usually squash the commits before merging? or just merge ?

rieder:
Depends on the PR, but in this case I think squash would be best...

ipelupessy:
@rieder are these ok? I think we determined that for examples style changes were sort of ok..

rieder:
> @rieder are these ok? I think we determined that for examples style changes were sort of ok..

These look fine to me.

Issue #175, opened 2017-10-26
async for overridden functions


ipelupessy:
the way asyncronuous calls currently work (as an attribute to the interface functions) has a minor drawback that it will fail for the (common) pattern where the function in question (often the evolve_model) is overridden...this could be fixed if implemented as a flag:
```
def evolve_model(tend, **kwargs):
   <some code>
   return self.overridden().evolve_model(tend, **kwargs)
```
with call: code.evolve_model(tend, async=True)
or is there some other way to do this??

ipelupessy:
fixed by #339

PR #339, opened 2018-11-14
multiple async requests on the same code

ipelupessy:
This patch allows for multiple async requests on the same code; see for examples
the new compile_tests/test_async.py

it also allows to specify async call with keyword argument async=True instead of the 
.async(..) method..

### DIFF
--- a/src/amuse/rfi/channel.py
+++ b/src/amuse/rfi/channel.py
@@ -57,8 +57,12 @@ class AbstractASyncRequest(object):
     def is_result_available(self):
         raise Exception("not implemented")
 
-    def result():
+    def result(self):
         raise Exception("not implemented")
+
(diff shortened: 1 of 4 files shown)


ipelupessy:
Note that this was already somewhat possible with ASyncRequestSequence..its now much easier:
```
instance=code()
instance.call1(.., async=True)
instance.call2(.., async=True)
instance.call3(.., async=True)
instance.async_request.wait()
```
with multiple codes you can combine them into a pool:
```
pool=instance1.async_request.join(instance2.async_request)
```

ipelupessy:
fixes also #175

ipelupessy:
async in this context can be used also in python3, however maybe change in the future..

PR #342, opened 2018-11-20
work on grid for samplePoints with explicit coordinate arrays

ipelupessy:
eg:
```
sample=grid.samplePoint(x=5, y=3)
```

 adds get_index, get_axes_names and possibility of specifying sample points with explicit coordinate vectors, also adds column-stack on quantities.


ipelupessy:
I changed the default method for the samplePoints, since interpolation only works for 3D atm, and nearest is probably fastest

PR #344, opened 2018-11-27
A bit of cleanup in LagrangianRadii.py

rieder:
e.g. use lengths_squared instead of custom function now, since it has been around for a while now.
Also use a proper main() function since it's cleaner that way (no globals).
Working on this file since there's a small problem with it - sometimes it returns LR(100%) but sometimes it doesn't. 
Should be more consistent.

### DIFF
--- a/src/amuse/ext/LagrangianRadii.py
+++ b/src/amuse/ext/LagrangianRadii.py
@@ -17,15 +17,12 @@ from amuse.ic.salpeter import new_salpeter_mass_distribution
 MassFraction = [0.005, 0.01, 0.02, 0.05, 0.1, 0.25, 0.5, 0.75, 0.9, 1.0] \
     | units.none
 
-# This could be more elegant in amuse as an attribute
-# for example as: stars.position.length_sq()
-# at the moment (April 2010) this is not yet implemented, but the alternative:
-# star.position.length()
-# is slow, due to the implicite square-root
-def distance_sq(stars, com) :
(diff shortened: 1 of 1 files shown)


ipelupessy:
ok, could you check the function in particle_attributes.py ?? (I prefer that one;-) )

rieder:
Thanks for pointing out the "other" LagrangianRadii. That one looks better (but also has some unnecessarily complex bits). Maybe we should add a deprecation warning of some kind to the one in ext?
In any case, the particles.LagrangianRadii method seems to give more consistent results than the method in ext (note that in several cases the results are not the same!). I'll use that one from now.

ipelupessy:
I think the main difference is the center of the cluster taken (ext version always uses the c.o.m., while 
the attributes version will default to searching the density center)

ipelupessy:
btw for the ext version: I don't think there is any reason to append L to integers??

ipelupessy:
that is surprising since default is reuse_hop=False...??

rieder:
Yes, very surprising. Took a while to figure out...

rieder:
I see now. Hop isn't stopped because it fails. 
Need to put in an exception to deal with this (and/or just always use some converter).

rieder:
> btw for the ext version: I don't think there is any reason to append L to integers??

Not sure what you mean?

ipelupessy:
nevermind, i misread...

Issue #345, opened 2018-11-28
Python3: 2to3 doesn't fix true-divs (Galaxy merger demo fails)


jhidding:
**Describe the bug**
In the file `src/amuse/community/galactics/interface.py`, line 315--318. There are several divisions by 7, which should be written as `x//7`. We don't know how many instances of this bug are elsewhere in the code.
Currently this code throws a `TypeError`: "TypeError: 'float' object cannot be interpreted as an integer"

**To Reproduce**
Run the [Galaxy merger demo](http://amusecode.org/doc/examples/applications/galaxy_merger.html) using Python 3.

**Expected behavior**
No exception being thrown.

**Environment (please complete the following information):**
 - OS and version: Debian Buster, Linux 4.18, Python 3.6.7.
 - Compiler: GCC 8.2.0

rieder:
Thanks for reporting. Made a pull request that fixes this, but I didn't check for other occurrences yet.

rieder:
Fixed by #346 (this specific instance)

PR #346, opened 2018-11-28
Make sure divisions are floor divisions

rieder:
Otherwise it will fail in Python3, see bug #345.
Need to check for possible other occurrences.


ipelupessy:
ok, we still need to check any other ocurrences, but I guess we will run into these at some point...

rieder:
This fixed #345

PR #348, opened 2018-11-29
Stop Hop on exception

rieder:
This should fix the hop stopping problem mentioned in #344


rieder:
This is not the cause after all.

ipelupessy:
I think a common exception raised at the add_particles can also be a MissingAttributesAmuseException, so I would at least print the original exception...

rieder:
Sure, that makes sense. I'll fix that.

PR #353, opened 2019-01-03
Patch code syntax (pep8 compliance)

rieder:
Whitespace changes, to comply more with AMUSE code style.
Changes were made automatically using autopep to fix errors/warnings of types E1,E2,E3,W2,W3 (https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes).
These are whitespace-only changes and will not affect how the code works, but they improve code readability.


ipelupessy:
hmm, I think we have discussed before that auto "fixes" tend to obscure code provenance, and are for that reason not preferred. Furthermore, I think will be annoying for people (such as me) who have uncomitted, wip, under development etc branches lying around...

rieder:
Hi Inti,

I understand the point about auto fixes not being preferred, but in this case I have taken time to understand exactly what the auto fixes are and I have also checked that they are done correctly.

Regarding the second point: git is good at solving such issues. It may still run into conflicts sometimes but everyone has development branches and there will always be times when these conflict somewhat.
In this case it's easy to fix by making the pep8 changes to your own branches too.

ipelupessy:
well, I don't to make life more complicated than necessary for people...will try some things first to see how bad it is...

goord:
Considering there are 41 forks out there, I think we should be very careful with these kind of massive syntax style changes. Lots of people will encounter merge conflicts, and I'm not sure whether everyone will know what to do about them...

rieder:
> Considering there are 41 forks out there, I think we should be very careful with these kind of massive syntax style changes. Lots of people will encounter merge conflicts, and I'm not sure whether everyone will know what to do about them...

Of course we should be careful, but that's not the same as "leave everything as it is". 
I believe most forks to be stale... Which is another problem.

ipelupessy:
@rieder: how did you implement the changes? (xo I can try your suggestion of applying the same changes to a branch...)

rieder:
@ipelupessy 
By running `autopep8 -i --select E1,E2,E3,W2,W3` on the .py files.

rieder:
Hi Inti,
Maybe we can have a look together on Friday or Monday, if either day works out?
I would think strategy 2 should work...

ipelupessy:
closing, given the potential for mayhem

PR #357, opened 2019-01-18
Fix bonsai

jbedorf:
This PR  fixes compilation of Bonsai1 and introduces  Bonsai2 to the master branch. 

Bonsai1:
- This disables deprecated and removed CUDA defines. In CUDA9 these references to 10+ year old hardware are no longer available which causes compilation to fail.

Bonsai2:
- Added a script that pulls a (working) version of Bonsai2 from GitHub, patches and builds the code.
- Includes unit-tests
- includes a README with short information on which shell settings to use in order to run it (`ulimit -s unlimited`).

Note: This is not the latest version of Bonsai2, the version here will be updated once Bonsai2 is compatible with CUDA9.2+


jbedorf:
Closing this PR it contains Bonsai2 data that should not be there.

jbedorf:
Reopened, now with proper Bonsai2 configuration and files.

jbedorf:
If you want to use messy defines, then yes. In practice I doubt much of that hardware is still alive...

ipelupessy:
is it good to merge? I agree with Steven it doesn't look too difficult to keep the support for older GPU alive...but maybe that should be done when/if someone asks for it..

rieder:
> is it good to merge? I agree with Steven it doesn't look too difficult to keep the support for older GPU alive...but maybe that should be done when/if someone asks for it..

Yes, I think that's fine (and Jeroen has convinced me that probably no one will since those GPUs are probably no longer used...).

jbedorf:
Yes. I do not plan to make any further changes to these particular versions anytime soon so might as well get it merged now.

Issue #310, opened 2018-09-26
Improve examples by not using "from amuse.lab import *" 


rieder:
Using "from amuse.lab import *" is easy, but makes it impossible to see where an imported object is defined. It would be better if we are explicit in importing objects, especially in the examples (which can be expected to serve as basis for new scripts).

The imports could still be from amuse.lab, but instead of the line 
```python
from amuse.lab import *
```
we could have e.g.
```python
from amuse.lab import (units, nbody_system, Hermite)
```

Or, perhaps preferably,
```python
from amuse.units import (units, nbody_system)
from amuse.community.hermite0.interface import Hermite
```

I'm happy to gradually modify examples if we agree this would be better.

ipelupessy:
+1

arjenve:
I can see two benefits to the current setup;

1. When changing the scripts to suit you needs, you do not have to learn / investigate the entire amuse structure to see from where a specific name needs to be imported
2. The 'from amuse.lab import *' decouples the public view of AMUSE from the internal structure, this we could use to change some of the internal structure without changing the external scripts.

I can also see a benefit of importing the specific module; this forces us to have a  less complex internal structure and make it clean and useful.

I do not want to change  'from amuse.lab import *' to 'from amuse.lab import Hermite' etc, it only increases the number of times a name has to be given. It also removes the ease of extending the script.

ipelupessy:
I am for clean&useful structure, and de-obfuscation! ;-)  I also agree that 
from amuse.community.hermite0.interface import Hermite is better than both versions of amuse.lab import...

rieder:
> 1. When changing the scripts to suit you needs, you do not have to learn / investigate the entire amuse structure to see from where a specific name needs to be imported

Sure, that's a benefit. But I don't think it's a major one, it 's easy enough to figure out where names are imported from. Especially if we use them in examples (as we do in some but not all currently).

> 2. The 'from amuse.lab import *' decouples the public view of AMUSE from the internal structure, this we could use to change some of the internal structure without changing the external scripts.

Except we don't change that structure? At least, not without simply deprecating the old structure while still keeping it working... So I don't see this as a problem for using specific imports.

> I do not want to change 'from amuse.lab import *' to 'from amuse.lab import Hermite' etc, it only increases the number of times a name has to be given. It also removes the ease of extending the script.

[...]

PR #358, opened 2019-01-19
Explicit module imports within examples/textbook

kjaehnig:
Commented out the use of "from amuse.lab import *" and added in individual imports of modules. Could only do textbook scripts that are supported on my laptop for testing.


rieder:
Hi @kjaehnig,
Thanks for contributing! we agree that the use of "import *" should be avoided.
However, in this case, the use of "from amuse.lab import *", is done on purpose, as the examples focus on source code rather than headers (see comments in #310).
I hope we can see further contributions from you!

Issue #362, opened 2019-01-23
Replace "print"s by python logging?


rieder:
In some places, AMUSE scripts (not just the examples) use "print" to log what is happening.
These could be replaced using the standard python "logging" module, which would make it easier to control the level (and direction) of output.
'logging' supports multiple output levels (such as "DEBUG", "INFO", and "WARNING"); various stream handlers (stdout, stderr, or a file); and formatted output which could also indicate the time and the module. All of this could be set in the user script.

E.g.: 
```python
print "this is an output at %s" % time.in_(units.Myr)
```
would become something like this:
```python
import logging

logger = logging.getLogger(__name__)

logger.info("this is an output at %s", time.in_(units.Myr))
```

As a secondary benefit, this would also get rid of needing print statement to print function conversion for Python3.

rieder:
note that strings passed to logging should be formatted, the following won't work:
```python
logger.info("This is an output at time", time, " for N particles", len(particles)")
```
Naively, it could be done with extra parentheses, but the following is better:
```python
logger.info(
    "This is an output at time %s for N particles %i",
    time, len(particles)
)
```

PR #363, opened 2019-01-23
Patch logging

rieder:
Attempt at implementing #362 (work very much in progress)

### DIFF
--- a/src/amuse/datamodel/grids.py
+++ b/src/amuse/datamodel/grids.py
@@ -1,3 +1,6 @@
+import logging
+import numpy
+
 from amuse.support.core import CompositeDictionary, late
 from amuse.units import constants
 from amuse.units import units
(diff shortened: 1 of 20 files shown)


ipelupessy:
good idea! following the "when to use logging" bit in https://docs.python.org/2/howto/logging.html
suggests keeping prints in the examples, and using logging/ warnings in the framework code
for things that report normal operation 

I think if there are obvious debug comments, you can also just comment them out...

rieder:
I agree, in examples using "print" would be preferable. Modules should use logging.
Especially multiples and encounters will require some work...

rieder:
What should be done with "verbose" user settings?
It would make sense to me to get rid of these entirely when we're using logging instead of printing.

rieder:
i.e. the "verbose=True" option in Bridge() etc.

ipelupessy:
I think the verbose option implies that you have control over the output of every instance, so you could make a logger at init:
```
class A(object):
  def __init__(self, verbose=False):
    self.l=logging.getLogger(str(self))
    if verbose:
        self.l.setLevel(logging.DEBUG)  
```

rieder:
Yes, that would make sense.

rieder:
(I did not yet apply this to the latest commit)

ipelupessy:
many conflicts...also: what is the scope of the PR? I think breaking up and doing PR with partial replacement is an option..

ipelupessy:
in view of the conflicts and comments this needs to be redone carefully, closing

Issue #305, opened 2018-09-17
Remove keyword collisions


rieder:
**Describe the bug**
amuse/rfi/core.py line 126 defines an "async" function. In Python 3.5+, async is a reserved keyword (https://www.python.org/dev/peps/pep-0492/). We may need to rename this function to prevent "invalid syntax" errors.

ipelupessy:
I don't think its a big problem since the async are class methods??
(there are other collisions in the code..)

ipelupessy:
we can maybe gather those here?

rieder:
> we can maybe gather those here?

good idea. I'll rename the issue.

ipelupessy:
new option for async as keyword argument to function calls (#339 )

ipelupessy:
ok, asyncawait was introduced in 3.5 and 3.6 but 3.7 makes it reserved..so will change

fjansson:
After discussing with Inti, I like the following:
* change the `async` function keyword to `return_request`
* change the `.async()` method to `.asynchronous()`
* encourage use of the function argument over the method since this allows wrapping the interface function

rieder:
sounds good to me.

ipelupessy:
async is fixed (4b48747)

ipelupessy:
other cases (not specific to python 3):
- object
- format
- callable

ipelupessy:
- set

rieder:
#359 is a duplicate of this issue

ipelupessy:
i think for "object" as used in the interface definitions (define_parameters, define_methods etc ) this can be replaced by "handler" (these instances of amuse.support.interface.HandleParameters etc)

rieder:
yes, that sounds good.

rieder:
* sum
(e.g. in src/amuse/ic/kingmodel.py)

PR #365, opened 2019-01-28
Replace 'object' with 'handler' where object was not a Python object

rieder:
Patch for #305, replacing 'object' with 'handler' in places where we should not be using 'object'.
Work in progress.

### DIFF
--- a/examples/textbook/helium_star.py
+++ b/examples/textbook/helium_star.py
@@ -92,7 +92,7 @@ def plot_tracks(temperatures_original, luminosities_original,
 
     save_file = 'HertzsprungRussel_HeliumStar.png'
     pyplot.savefig(save_file)
-    print '\nSaved figure in file', save_file,'\n'
+    print('\nSaved figure in file', save_file,'\n')
     pyplot.show()
 
 if __name__ == "__main__":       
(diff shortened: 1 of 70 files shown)


rieder:
Any more places where this should be replaced?

ipelupessy:
I think these pattern should be safe to replace:
define_.*object.* 
object.add_.*
object.ERROR_CODE
object.INDEX
object.NO_UNIT

also: some logging wip seems to have slipped in

ipelupessy:
the travis tests do not yet cover the code interfaces..you have to test this locally

rieder:
> I think there a little bit too many replacements, cases when object occurs in the class definition ( and maybe other uses) should not be replaced, ie
> 
> ```
> class FastKickDoc(handler):
> ```
> is wrong.

This one slipped through, but I think I got them elsewhere... I tried to pay attention to that.

rieder:
> also: some logging wip seems to have slipped in

Ah, yes. Was clearly not the intention, I'll undo that.

ipelupessy:
think its good to go..any errors or typos should surface quickly..

rieder:
Agreed. I will merge it then.

PR #370, opened 2019-02-06
Fix compatibility issue for python3

rieder:


ipelupessy:
this should not be necessary (auto converted by 2to3), if you want to remove the prionts, they should be warnings or logged as warnings #363 !!

rieder:
It should be, but it wasn’t automatically converted for some reason. But yes, it should indeed be logged and not printed. Will change that.

PR #378, opened 2019-02-19
Cleanup of setup.py, add missing requirements

rieder:


rieder:
Installing required python packages is now easier:
pip install -r requirements.txt

rieder:
Change the version number to 12.0.a0 (alpha version, should be pip-installable)

rieder:
Also cleanup of support/setup_codes.py.

ipelupessy:
- here, it doesn't accept the version number, changing it to 12.0a0, in any case I don't think its necessary to label it alpha..(is there a convention for alpha labeling?)
- why/ when does it need wheel?
- for the syntax:  you using the same autopep options as in the previous PR (autopep8 -i --select E1,E2,E3,W2,W3)? its still painful to scan/review a PR with a mix of "real" changes and 
cleanup... 
- the long description turns up on pypi? Its a bit weird to read there because it has a lot of install and build instructions

rieder:
- The version on PyPI is this PR plus a few changes that I did not yet commit (which includes the version numbering which I changed precisely because this one did not work).
- I'm not sure if wheel is strictly needed tbh. I simply do not have an installation without it (virtualenvs automatically install it), so it's difficult to test. 
- Syntax: I was more selective (and manual) with this PR, but it should not be very different than that. Sorry for mixing the two. The "real" changes here are additions to setup() (mpi4py and h5py as requirements, description) and removing unused imports (and sorting the others).
- Indeed, the long description should be cleaned up...

ipelupessy:
any idea what the problem with python3 is?

rieder:
Not yet. It did not build, and I haven't had time yet to check why.

ipelupessy:
python3 worked here...

ipelupessy:
merge? maybe should reinstate python3?

rieder:
This PR is a bit tricky to merge so I'm closing it and re-applying changes in a new PR.

rieder:
Wheel is definitely necessary to build for Python3. Without, it failed to install for me.

Issue #163, opened 2017-09-21
Make installing AMUSE easier (maybe with pip?)


rieder:
If AMUSE could be installed with a simple 'pip install amuse', that would make it much more accessible.
https://packaging.python.org/tutorials/distributing-packages/ details the necessities for this. 

Issues this depends on (incomplete):
#144 #133

rieder:
I'm not at all sure if it will be possible to install AMUSE fully in this way, especially taking into account the community codes. However, it should be possible for the 'core' (minimal) AMUSE.

rieder:
Relates to #186

ipelupessy:
I have added this to team sprint project

egpbos:
Now that we can build and run almost all of Amuse on macOS using conda (#235), I think the last step of actually making Amuse a conda package (also for Linux) would be relatively easy.

The choice between conda and pip depends a bit on the progress on decoupling Amuse's core from the community packages. Amuse core depends almost exclusively on Python packages (see https://github.com/amusecode/amuse/blob/master/.travis.yml#L30; only exception is the Fortran compiler, which can probably easily be taken out), so for Amuse core, pip would be sufficient. Those Python packages may in turn depend on libraries that still need to be installed, but that is probably manageable. However, if people want to install the whole Amuse package, they will need some non-Python packages. These are available on conda, but not on pip.

egpbos:
So related to #107 :)

rieder:
Hmm, the real problem here is a line in support/create_codes.py that overwrites the data_directory with something relative to '/' (line 115). This fails. Line 115 overwrites 'data_dir' set in line 114, which is the correct way of setting it. Removing line 115 makes pip install AMUSE.

rieder:
I managed to upload & install a slightly modified (minus a few codes) version of AMUSE to PyPI, and install it via pip:
https://pypi.org/project/amuse/

ipelupessy:
hmm, apperently line 115 is much older than I thought....

ipelupessy:
I think the problem is that data_dir in the run of GenerateInstallIni is used to fill in configuration of amuserc, assuming the --root will be the "/" of the install target (ie build for a package manager), while in the run of  InstallLibraries its used in the script to copy stuff to...

rieder:
I uploaded a new version to PyPI, see branch 
https://github.com/rieder/amuse/tree/release-12.0a3 for the source.
[...]

PR #379, opened 2019-02-19
Remove incorrect setting for data_path which overwrites the correct one

rieder:
This is needed for installing AMUSE via Pip (which fails on this line).
Apologies for branching on the main AMUSE repository...


ipelupessy:
well...I am trying to remember where this line came from (its seems its a dirty patch for some issue which I forgot to properly fix...)...its ok to merge this though, and fix it properly next time I encounter it...

ipelupessy:
see the comment on #163

ipelupessy:
conclusion is that yes, the datadir here should not be relative to "/" ..

ipelupessy:
I think it was still wrong, the datadir was made into a relative path of the build dir, changed it (maybe check?)

PR #388, opened 2019-03-06
fixes for determination of amuse_root_dir

ipelupessy:
..and fix for amusrc in case of pip install


ipelupessy:
its a bit of a mess: amuse root dir sometimes indicates location of build.py and data dir,
other times needed as root of amuse src tree (the import location..)

PR #389, opened 2019-03-07
Move gencode

ipelupessy:
this moves gencode (the generator for interface code) into the amuse module tree, and removes the support directory from the install (I don't think anything there has to be included)


rieder:
Looks fine to me.

Issue #132, opened 2017-05-23
Flatter AMUSE namespace?


rieder:
The AMUSE namespace is not all that easy to remember at the moment, and different from what is used in other major projects.

For example:
```python
# Expected:
import amuse.units
x = 1 | units.cm
# this does not work

# Actual:
from amuse.units import units
x = 1 | units.cm
```

and:
```python
from amuse.datamodel import Particles
```
(which I always forget)

This leads to the use of things like 
```python
from amuse.lab import *
```
which are not at all ideal... 

I would suggest we try to make the AMUSE namespace somewhat flatter, in the way that projects like astropy/numpy/scipy and others work. 

E.g.:
```python
# import everything under amuse namespace
import amuse 
# or, for a shorter namespace:
import amuse as am

particle = amuse.Particle()
particle.x = 1 | amuse.units.cm
```
```python
# imports units directly
import amuse.units 
x = 1 | units.cm
```

Similarly, we could then have (these are suggestions)
```python
# import particles
import amuse.particles 
```
```python
# import all interfaces
import amuse.interfaces
# or only one
import amuse.interfaces.hermite0 as Hermite
```

ipelupessy:
hmm, I don't think things like:
```
  from amuse.units
  d=1| units.m
```
are supposed to work. I think 
```
  from amuse import units
  d=1|units.m
```
could be nice, but then you would also want:
```
  from amuse import constants
```
not sure if this will fit...in any case you would only want changes that don't break existing
scripts...as a side node, I consider things like
```
  import numpy as np
  import amuse as am
```
bad form, don't do it...

rieder:
Regarding the latter point: in the case of numpy, they use "np" _everywhere_ in their documentation...

ipelupessy:
maybe something like: https://github.com/ipelupessy/amuse/tree/simplified_namespaces

 I am not sure if it will work 100%, and in any case in general I am not in favour of obfuscation of namespace, in case you didn't notice;-)

arjenve:
Yes, moving all that is now in lab to the amuse main __init__ would be a good idea, and allowing
for "from amuse import *" .  
I do not want the renaming to am in our examples like numpy always seems to do in their examples, (why not call the package np?). 
But yes, having from amuse import <name> is helpful (btw we already do this for datamodel where we import particles etc into the datamodel __init__)

arjenve:
BTW instead of interfaces, I would name these the codes.

rieder:
[...]

PR #390, opened 2019-03-07
Packages tryout

ipelupessy:
this is just for testing atm


rieder:
amuse-sse should depend on amuse-framework being installed.

rieder:
do we call it amuse-framework? amuse can then be a meta-package that does a full installation.

ipelupessy:
yea, I didn't pay attention to names and/or dependencies yet...(your previous proposal was ok)
I think all codes and packages after the framework can use the same support/setup_codes.py so it may be an idea to move this (or join if possible) to the root support...the packages will need different setup.py (but they still can be simplified /slimmed down further...)

ipelupessy:
btw have you seen #391 ??

ipelupessy:
I think we can remove distributed from the framework package (since also it is not working atm)
also, I think it may be nice to add a function that reports the available codes for an installation...

rieder:
Yes to both. 

By the way, why is distributed not working? Is there an issue for this?

ipelupessy:
there is a failure to communicate with a Pilot process (when using ssh starting Pilot on another machine) - I am checking whether I misconfigured something, but I think not...(will make a proper issue otherwise)

ipelupessy:
i have made some changes (#395)..these are merged here too, simplifies the package directories...may neeed some testing

rieder:
I'm testing this implementation now for the Phantom interface (https://github.com/rieder/amuse/tree/patch-add-phantom). Seems to work quite well. I can now install amuse-phantom with only amuse-framework as requirement, saving a lot of time on compiling packages that shouldn't need recompiling when I change a small thing in the Phantom interface.
Only "issue" is amuse.lab - that won't work until all amuse 'base' community packages are installed. But I don't think that is a problem. It can be solved by installing a meta-package 'amuse' which depends on all of these.
I think this method is the way forward for installing AMUSE.

ipelupessy:
I agree;-) The only task left is maybe generating all the subpackages - I think this could be done semi-automatically (and add this to the 'amusifier' script #397 ??)

rieder:
Let's get this done when I visit Leiden after easter?

ipelupessy:
sounds good..

rieder:
We may want to rethink the "DOWNLOAD_ONLY" codes as well in this process. 
If these are separate packages, this flag makes no sense anymore. Probably we should exclude them from the "amuse" meta-package instead, and not include them in lab.py?

ipelupessy:
think lab should detect which codes are available...

rieder:
Yes that would be a good option. If a code is not installed (i.e. ImportError is raised), we could show a message with installation instructions (e.g. 'pip install amuse-${codename}).

ipelupessy:
for the imports in lab: something like this is actually in codes.py... (looks like *old* code)

rieder:
That file was last touched 9 years ago, so old indeed!
But yes, the general idea of that file seems like a good one to implement once again. 
I'll have a go at it.

rieder:
Maybe we should have a go at #132 while we're at this... 
And perhaps also a more unified way of naming codes/interfaces (currently it's anyone's guess which letters are capitalised). E.g.: hermite0/HermiteInterface -> hermite/hermiteInterface; phiGRAPE/PhiGRAPEInterface -> phigrape/phigrapeInterface
(move to consistent names and allowing the current names to work with a deprecation warning)

rieder:
I can't figure out how to do a "from X import Y" using importlib and import_module... @ipelupessy do you know how this could work?

ipelupessy:
you mean:
```
def get_code_interface(code, classes):
  module=importlib.import_module("amuse.community."+code+".interface)
  for class in classes:  
    globals[class]=getattr(module, class)
```
called as:
```
get_code_interface("fi", ["FiInterface", "Fi"])
```
???

ipelupessy:
btw, i have changed some things simplifying the setup.pys..hopefully it didn't break anything (it shouldn't..)

rieder:
> you mean:
> 
> ```
> def get_code_interface(code, classes):
>   module=importlib.import_module("amuse.community."+code+".interface)
>   for class in classes:  
>     globals[class]=getattr(module, class)
> ```
> called as:
> 
> ```
> get_code_interface("fi", ["FiInterface", "Fi"])
> ```
> ???

Ah yes, that looks good.

rieder:
> btw, i have changed some things simplifying the setup.pys..hopefully it didn't break anything (it shouldn't..)

I'll check (but first impression is good - much cleaner now!)

ipelupessy:
its not working yet (minor import problem)..will fix it in a bit, catching train...

ipelupessy:
should be fixed now..

ipelupessy:
still broken..

PR #394, opened 2019-03-21
Savepoint on incode particles

ipelupessy:
this patch allows code particle sets to have a savepoint.
This savepoint is saved in memory (like ordinary).


ipelupessy:
a bit simple minded maybe, but I think it works...
(btw, first commit is unrelated, already in master)

rieder:
Looks fine to me, but it would be better if we can catch a specific exception here...

ipelupessy:
I leave like it is now - the same is done in grids.py; I may want to change it to do the copy for all cases (i don't see a good reason why that shouldn't always work...)

PR #397, opened 2019-03-25
make config.py read its info from config.mk, and introduce

ipelupessy:
amuseifier as installed build.py script

amusifier also has an option to output AMUSE_DIR


ipelupessy:
not sure what the best spelling is for amus(e)ifier

PR #408, opened 2019-05-02
add package for tests

ipelupessy:
this package should include all the tests

user can then:
nosetests amuse.test.suite

issues:
currently it has amuse as prereq. maybe better idea to allow partial installs to be tested (or include tests in the respective packages...)


ipelupessy:
now it depends on amuse-framework and installs all tests - I think that is ok..

main issue is the fact that it generates a lot of test output and build files in the cwd..not nice..

rieder:
Do the tests really need to create permanent output?  Maybe just put it in a tempdir?

rieder:
I think this package is fine for now, but I think in the long term it would be better to package code-tests with their respective code.

rieder:
Still, there will be a need for non-code-specific tests and tests that depend on multiple codes, so this package should exist.

ipelupessy:
yea, I think for now its ok. its use would be for debugging: if an issue crops up you can just install all test and run only the ones you need..

rieder:
Ok - I've uploaded it to PyPI now.

PR #409, opened 2019-05-02
Patch consistent naming

### LINKED ISSUE #399: Guidelines for naming community packages / modules
Currently, community code modules (and their functions) aren't named consistently.
E.g. a code 'hermite0' has a function 'HermiteInterface'; 'bhtree' has 'BHTreeInterface'; 'phiGRAPE' has 'PhiGRAPEInterface'.
It would be preferable to have consistent naming for all of these, especially now that we are moving to a modularly installable AMUSE.

**Describe the solution you'd like**
I propose that:

- modules are always named with lowercase only (there are very few exceptions anyway)

- the functions should be in either UpperCamelCase or lowerCamelCase (Upper would be most in line with the current situation but lower looks better in my opinion). No additional capitalisation should be used, even if this is the "proper" name of the code (e.g. 'BHTreeInterface' would become 'BhtreeInterface' or 'bhtreeInterface').

- the name of the module is used exactly in the functions they provide (e.g. 'hermite0' should have 'Hermite0Interface'/'hermite0Interface', *or* 'hermite' should have 'HermiteInterface'/'hermiteInterface'. If additional functions are provided (e.g. in 'fi'), they also use this name followed by an additional description (e.g. 'FiMapInterface'/'fiMapInterface').

- numbers in the name should be avoided, unless we (expect to) support multiple concurrent (and incompatible?) versions of that code

- Current modules and functions not in line with this should import the new names, while displaying a deprecation warning. I see no harm in keeping these names around for the foreseeable future, but we should encourage the stricter naming and not accept new codes not adhering to this.

Related to this would be the naming of packages on PyPI, for which I'm not yet sure what is the best naming scheme (consider possible packages could be from 'community' as well as 'ext', and possibly other *muse projects).
Simplest would be e.g. 'amuse-bhtree', but other options could be 'muse-amuse-bhtree', 'amuse-community-bhtree' and things I did not yet consider...

I'd welcome comments to all of this (@ipelupessy @stevemcmillan @spzwart)

rieder:
Attempt to fix #399 
(work in progress)


rieder:
I think I renamed all occurrences of hermite0 and phiGRAPE (outside of the docs).

ipelupessy:
looks good, some weirdness (probably github related:
I got the following rename listed:
 src/amuse/community/hermite0/interface.h → src/amuse/community/hermite/__init__.py

furthermore, I would announce this change on the mailing list...

rieder:
huh, that is weird. I will double check.

I agree, this change should be announced.

rieder:
My guess is that it's caused by both files being empty, and apparently git or github finds it difficult to determine which is renamed which. The result seems fine.

ipelupessy:
for me, this one can be merged, maybe announce the change in the 12.0 release?

rieder:
Ok!
I will merge it.

ipelupessy:
for mmams maybe we should add or change:
MMAMS=MakeMeAMassiveStar
MMAMSInterface=MakeMeAMassiveStarInterface

rieder:
I'm in favour of adding the new full name.

rieder:
Maybe also rename the ph4 interface PH4? @stevemcmillan?

rieder:
> I'm in favour of adding the new full name.

Of course, I had this reversed. The full name is the current one, the shorter is the consistent one! Added now.

PR #413, opened 2019-05-07
Updates to prerequisites install script

### LINKED ISSUE #411: Add script to create virtualenv and install prerequisites
It would be beneficial to have a script that generates a virtualenv, installs prerequisites (e.g. (Open)MPI etc), and installs AMUSE (with pip).
This would be similar to the original way of installing from scratch, but within a virtual environment and without having to build Python and its modules.

**Describe the solution you'd like**
Create an "install-amuse.sh" script that takes a directory as argument (and possibly checks for already-installed prerequisites).

rieder:
PR to fix #411.
Work in progress.

### DIFF
--- a/doc/install/install3.py
+++ b/doc/install/install3.py
@@ -1,7 +1,5 @@
 #!/usr/bin/env python
 
-
-
 import sys
 import os.path
 import os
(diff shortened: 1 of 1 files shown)


ipelupessy:
if it works, can it be merged? (even if incomplete)

rieder:
yes, should be fine (please test). but should probably also update the install.py script similarly.

PR #415, opened 2019-05-08
test of travis with pip install of amuse

ipelupessy:
this is a very naieve test..wip


ipelupessy:
clearly not the way to do it

PR #418, opened 2019-05-09
add option to abort setup (or pip) in case build error detected

ipelupessy:
add an option to stop build (and hence a pip install) in case of build error.

It may need some refinement (i can see at least 3 reasonable options:
- never abort
- abort if any build failed
- abort only if all builds failed
)

### DIFF
--- a/support/__init__.py
+++ b/support/__init__.py
@@ -1,4 +1,4 @@
-supportrc=dict(framework_install=True, package_name="amuse")
+supportrc=dict(framework_install=True, package_name="amuse", abort_on_build_failure=True)
 
 def use(arg):
     if arg == "package":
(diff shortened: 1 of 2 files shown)


ipelupessy:
so this option can be set individually in the packages' setup.py like:
```
from support import set_abort_on_build_failure
set_abort_on_build_failure(True)
```

rieder:
Yeah that would be handy. 
I’ve seen packages install that actually failed to build the worker, would be good to avoid that.

ipelupessy:
I will think a bit about it...
(environment variable control would be another option)

ipelupessy:
its now controlled by allow_build_failures with 3 options none, some or all which allows
either no failures, some (= continue as long as 1 build succeeded, also the default) or all (continue regardless) still controlled with supportrc (on the package level) or with a environment variable override
(AMUSE_ALLOW_BUILD_FAILURES)

PR #434, opened 2019-06-19
hide lab import failures of community codes

ipelupessy:
make a place holder class with error message instead


ipelupessy:
hermite and phigrape renames are anticipated upon, MMAMS needs to be fixed

rieder:
MMAMS rename fixed

PR #435, opened 2019-06-21
add phigrape package to amuse package

### LINKED ISSUE #432: add python cycler to list of prerequisites
Many codes and examples fail if the python module 'cycler' is not installed. 

It is used through matplotlib, but apparently not required by it, at least, when installing matplotlib from rpm on CentOS 7, it isn't included. This may be due to matplotlib versions, in Fedora 29 and in Ubuntu 19.04, cycler is installed as dependency of matplotlib.
Perhaps with the move to the pip-based install of amuse, it could be listed as dependency of one of the packages.

rieder:
Also fix #432 by requiring matplotlib (>=2.2, which still supports Python 2.7)
Also change amuse package classifiers (don't support 2.6 anymore since matplotlib doesn't, amuse isn't beta anymore, more specific on which Python 3 versions we know to work)


ipelupessy:
its ok, though I am not super convinced by requiring matplotlib (bit large prereq. and only necessary for the examples, maybe we should have a seperate examples package)...

ipelupessy:
I don't think the examples are actually anywhere in the packages?? so no reason to require matplotlib...

rieder:
Perhaps. But if you want a lightweight installation you probably don’t want to use the amuse package anyway...

ipelupessy:
merge?

computerworkers:
Hello amuse team,
many of the `textbook' examples do not work because they fail with

``from amuse.community.phiGRAPE.interface import PhiGRAPE, PhiGRAPEInterface
ImportError: No module named phiGRAPE.interface``

However when you do (note the letters case)
``from amuse.community.phigrape.interface import phigrape`` you get no errors. Could you please check all examples?

rieder:
This is due to a very recent change (#409), though only the module name should be lowercase (I need to verify this). 
I thought the examples were properly modified at the same time, but I’ll verify that as well. 

This change will be announced shortly on the mailing list, with the release of Amuse v12.

Thanks for reporting!

rieder:
Merging.
Fixes #432 as well.

PR #442, opened 2019-06-27
new approach to fix missing code imports

ipelupessy:


ipelupessy:
this has the potential to mess up builds...

ipelupessy:
i think it should now work okish..the only wonky thing is with the cython build right now

ipelupessy:
I am closing this, the most urgent problems are fixed by #473

Issue #451, opened 2019-07-08
replace c++ binding (datatypes only?) in higpus


ipelupessy:
higpus has wat looks as some remnants of c++ mpi bindings (datatypes like MPI::BOOL) - these should be fixed (ideally on some machine where gpu stuff can be tested)

rieder:
@mariosperaastro can you creatie a pull request for a fix?

mariosperaastro:
@rieder I'll do that!

rieder:
Fixed by #455

PR #455, opened 2019-07-16
Issue fix

mariosperaastro:
Fixes issue #451 and adds other minor mods.

### DIFF
--- a/src/amuse/community/higpus/interface.cc
+++ b/src/amuse/community/higpus/interface.cc
@@ -31,7 +31,7 @@ map<int, dynamics_state> dm_states;
 map<int,dynamics_state>::iterator it;
 map<int,dynamics_state>::iterator et;
 
-int size, rank, particle_id_counter;
+int size_higpus, rank_higpus, particle_id_counter;
 unsigned int         N, M, NGPU, TPB, FMAX, BFMAX, MAXDIM, GPUMINTHREADS;
 bool                 CDM, CDV, VIR = 0, warm_start = 0, setdev = 0, cleanstop = 0, init = 0;
 double               EPS, ETA6, ETA4, GTIME, GTW, TTIME, ATIME, DTMAX, DTMIN, DTPRINT, plummer_core = 0.0, plummer_mass = 0.0, ratio = 0.0, mscale = 0.0, rscale = 0.0;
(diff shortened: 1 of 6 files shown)


src/amuse/community/higpus/interface.cc
```diff
@@ -193,10 +193,10 @@ int evolve_model(double t){
    FMAX = 1000000 + (int) ((GTIME+GTW) / DTPRINT);
 	
    TTIME=t;
-   HostSafeCall(Hermite6th(TTIME, &GTIME, &ATIME, local_time, step, N, M, pos_PH, vel_PH, pos_CH, vel_CH, a_H0, MAXDIM, NGPU, TPB, rank, size, BFMAX, ETA6, ETA4, DTMAX, DTMIN, EPS, DTPRINT, FMAX, warm_start, GTW, GPUMINTHREADS, plummer_core, plummer_mass, rscale, mscale, dev, &cleanstop, path));
+   HostSafeCall(Hermite6th(TTIME, &GTIME, &ATIME, local_time, step, N, M, pos_PH, vel_PH, pos_CH, vel_CH, a_H0, MAXDIM, NGPU, TPB, rank_higpus, size_higpus, BFMAX, ETA6, ETA4, DTMAX, DTMIN, EPS, DTPRINT, FMAX, warm_start, GTW, GPUMINTHREADS, plummer_core, plummer_mass, rscale, mscale, dev, &cleanstop, path));
 
    MPISafeCall(MPI_Barrier(MPI_COMM_WORLD));
-   MPISafeCall(MPI_Bcast(&cleanstop, 1, MPI::BOOL, 0, MPI_COMM_WORLD));
+   MPISafeCall(MPI_Bcast(&cleanstop, 1, MPI_C_BOOL, 0, MPI_COMM_WORLD));
```
ipelupessy: Shouldnt this (and the other MPI::BOOLs) be a MPI_CXX_BOOL ? It may not matter in practice, but that depends on compiler implementation of c and c++ bools

mariosperaastro: I think the MPI_CXX_BOOL type has been introduced very recently, and it depends on MPI 3.0. That's why I preferred to not use it. But yea, as you said, it may not matter in practice, especially for MPI_Bcast functions.

ipelupessy: ok, didn't realize it is that recent...

mariosperaastro: Actually, I think they introduced these CXX types in MPI-2.2... but then MPI-2.2 was also the last version to support CXX bindings. But then, MPI-3.0 re-introduced full support for some CXX datatypes...so, better to stick with C bindings, I guess :-D

rieder:
OK to merge?

ipelupessy:
it looks good, but my cuda compiler is out of date..not been able to test it yet..

PR #457, opened 2019-07-23
implement asynchronuous access to data stores

ipelupessy:
for example:

to access density of a grid of a code:
```
request=code.grid.request.density
dens=request.result()
```

some more testing and imlementation of setters is needed..

### DIFF
--- a/src/amuse/datamodel/base.py
+++ b/src/amuse/datamodel/base.py
@@ -731,6 +731,7 @@ class AbstractSet(object):
             
         object.__setattr__(self, "_derived_attributes", CompositeDictionary(derived_attributes))
         object.__setattr__(self, "_private", PrivateProperties())
+        object.__setattr__(self, "_request", AsynchronuousAccessToSet(self))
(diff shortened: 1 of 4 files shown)


ipelupessy:
@rieder this PR introduces a special attribute request on data stores which can be used to set and request attributes  asynchronuously, e.g.
request=code.particles.request.vx 
return inmediately with a request for vx:
vx=request.result()
waits and blocks for vx to become available
setting:
code.particles.vx=..
will continue w/o blocking. There is no return request. You can wait on;
code.async_request.wait() 
for the call to finish, but normally there should be no reason to do so
(any following call will either wait for it to finish, or result in a queued request)
so, may question: do you think this way of implemting is ok? alternative could be:
code.particles(return_request=True).x
or making special attributed prepending something like: 
code.particles.request__vx
??

rieder:
@ipelupessy I think I like your suggested method better than the alternatives. It seems clear enough.

PR #460, opened 2019-07-29
Make the amuse namespace less deep

### LINKED ISSUE #132: Flatter AMUSE namespace?

rieder:
Attempt to fix #132 (work in progress)

### DIFF
--- a/src/amuse/__init__.py
+++ b/src/amuse/__init__.py
@@ -1,22 +1,22 @@
 """
 The Astrophysical Multipurpose Software Environment
 
-The aim of AMUSE is to provide a software framework, in which existing codes for 
-dynamics, stellar evolution, hydrodynamics and radiative transfer can easily be 
-coupled, in order to perform state-of-the-art simulations of a wide range of 
+The aim of AMUSE is to provide a software framework, in which existing codes for
+dynamics, stellar evolution, hydrodynamics and radiative transfer can easily be
+coupled, in order to perform state-of-the-art simulations of a wide range of
 different astrophysical phenomena.
 
 It contains several packages, most notably:
-units     - AMUSE uses quantities, i.e. a number (or array) with a unit attached, 
+units     - AMUSE uses quantities, i.e. a number (or array) with a unit attached,
             instead of just numbers (vital when coupling different codes!)
 datamodel - defines particles and grids, on which data (quantities) can be stored
 ic        - a collection of routines to generate initial conditions
 io        - how to read and write data in several file formats
-community - a variety of existing codes of different physical domains, each with 
+community - a variety of existing codes of different physical domains, each with
             a uniform interface to enable coupling
 
 Help is available for each of these packages, e.g.:
-> amuse.sh
+> python
 >>> import amuse.ic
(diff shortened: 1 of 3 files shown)


src/amuse/__init__.py
```diff
@@ -27,12 +27,34 @@
...
+from amuse.units import (
+    units, nbody_system, constants, generic_unit_system,
+)
+from amuse.units.quantities import (
+    ScalarQuantity, VectorQuantity, AdaptingVectorQuantity,
+    zero
+)
+from amuse.datamodel import (
+    Particle, Particles, ParticlesSuperset, Grid,
+    particle_attributes,
```
ipelupessy: why particle_attributes? In general I don't think
```
from amuse.datamodel import Particles
```
etc is so bad (a lot of submodules already imported into datamodel..).. same goes for amuse.io

ipelupessy:
maybe cut the codes namesspace furhter short to:
```
from amuse.community import BHTree 
```
??

ipelupessy:
although this leaves the same problem as in lab with the separate packages...

rieder:
Indeed. I thought about this, but amuse.community.codename seems clear enough since the codename is repeated as the interface name. 
‘’’python
from amuse.codes import CodeName
‘’’
is another option (an already existing if ignored/obsolete one at that).

I think the aim should be to reduce the need for “from amuse.lab import *” type imports by making it obvious how to import the necessary bits for a script. Cutting out the “.interface” bit helps enough I think, if it is obvious that the “community” namespace is where all the codes are.

src/amuse/__init__.py
rieder: mainly because I was copying from lab.py... particle_attributes probably shouldn't be imported here (I can't recall using it).

you're right, it's not so bad - but still, having the "most essential" bits available in the amuse namespace can help new users (and hopefully prevent bad style imports).

ipelupessy:
btw, it might be good to remember here and for similar changes non astro use of AMUSE...

rieder:
@stevemcmillan had some thoughts on this as well, we should discuss it this week.

rieder:
Closing this PR since we selected an alternative approach

Issue #232, opened 2018-06-11
add CFF (citation file format) file


ipelupessy:
**Is your feature request related to a problem? Please describe.**
its related to citation for the AMUSE

**Describe the solution you'd like**
add CFF file (citation file format file: https://github.com/citation-file-format/citation-file-format) with the 
references for AMUSE code papers. (not references for the community codes)

**Describe alternatives you've considered**
The default citations only show up after using AMUSE. CFF files are also used by automated tools to generate citations

**Additional context**
CFF are an emerging standard for software citations. I am no particular fan of the format itself (doesn't look very elegant), but if this is what other people are using...

rieder:
Seems good to have something like this.

Perhaps we could additionally generate bibtex files with ADS entries?
We could easily get those for all the citations that are currently printed.

ipelupessy:
fixed by #311

PR #469, opened 2019-09-10
Add ADS bibcodes to interfaces

rieder:
Work in progress to add ADS bibcodes to all literature references.

### DIFF
--- a/src/amuse/community/aarsethzare/interface.py
+++ b/src/amuse/community/aarsethzare/interface.py
@@ -7,8 +7,8 @@ class AarsethZareInterface(CodeInterface, LiteratureReferencesMixIn):
     Aarseth & Zare
 
     The relevant references are:
-        .. [#] Aarseth, S. & Zare, K., 1974, Celestial Mechanics 10, 185.
-        .. [#] Aarseth, S. & Zare, K., 1974, Celestial Mechanics 10, 516.
+        .. [#] Aarseth, S. & Zare, K., 1974, Celestial Mechanics 10, 185 [1974CeMec..10..185A]
+        .. [#] Aarseth, S. & Zare, K., 1974, Celestial Mechanics 10, 516 [1974CeMec..10..516A]
     """    
     #include_headers = ['worker_code.h']
 
--- a/src/amuse/community/athena/interface.py
+++ b/src/amuse/community/athena/interface.py
@@ -34,12 +34,12 @@ class AthenaInterface(CodeInterface, MagnetohydrodynamicsInterface, LiteratureRe
     different Riemann solvers and spatial reconstruction methods.
     
     The relevant references are:
-        .. [#] Gardiner & Stone 2005, JCP, 205, 509  (2D JCP Method)
-        .. [#] Gardiner & Stone 2007, JCP, 227, 4123 (3D JCP Method)
-        .. [#] Stone et al. 2008, ApJS, 178, 137 (Method)
-        .. [#] Stone & Gardiner 2009, NewA, 14, 139 (van Leer Integrator)
-        .. [#] Skinner & Ostriker 2010, ApJ, 188, 290 (Cylindrical Integrator)
-        .. [#] Stone & Gardiner 2010, ApJS, 189, 142 (Shearing Box Method)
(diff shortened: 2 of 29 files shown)


rieder:
This PR aims to improve the literature references in community codes by
1. adding ADS bibcodes to all the references for easier lookup, in the format [bibcode].
2. presenting the references in a common format (to be decided what that should be)

spzwart:
Yes, that is a good idea. I always wanted to make that change, but never got around doing it.
Could we make a central data-repository with all the references and have the read-in by the respective codes, or is it better to keep the ref in the appropriate interface file, where it currently is?

In term of content: what about the ADS bibtex code and the CMDA reference style (which is more complete than the A&A or MNRAS styles.
This PR aims to improve the literature references in community codes by

rieder:
For (2), I am thinking of the following:

```
You have used the following codes, which contain literature references:

	"CodeInterface"
		Author A., Buthor B., [docstring]:
		... "Article Title"
		** Cuthor C., Duthor D., [docstring]:
		... "Article Two Title"
		... Additional comments for Article Two
```

rieder:
Also, there should be literature references for the use of Python and Numpy, and (if used) Matplotlib. And probably some others.

rieder:
And some explanation for the asterisks (** means "citing these articles is not mandatory but recommended", no asterisks means "need to cite these", one asterisk is not defined?)

spzwart:
I was thinking of:

2018araa.book.....P: Portegies Zwart, S., McMillan, S. 2018. Astrophysical Recipes; The art of AMUSE.Astrophysical Recipes; The art of AMUSE.
or
Portegies Zwart, S., McMillan, S. 2018. Astrophysical Recipes; The art of AMUSE.Astrophysical Recipes; The art of AMUSE ( 2018araa.book.....P)

rieder:
> Could we make a central data-repository with all the references and have the read-in by the respective codes, or is it better to keep the ref in the appropriate interface file, where it currently is?

I prefer to keep it local, since we are making AMUSE more modular and a centralised citation library would not help with that.

> In term of content: what about the ADS bibtex code and the CMDA reference style (which is more complete than the A&A or MNRAS styles.

> I was thinking of: 2018araa.book.....P: Portegies Zwart, S., McMillan, S. 2018. Astrophysical Recipes; The art of AMUSE.Astrophysical Recipes; The art of AMUSE. or Portegies Zwart, S., McMillan, S. 2018. Astrophysical Recipes; The art of AMUSE.Astrophysical Recipes; The art of AMUSE ( 2018araa.book.....P)

That's fine with me, but I think I would prefer multiple lines for increased readability. I think I would favour the docstring at the beginning of the line (easier to find, since it doesn't move around).

rieder:
We could move the references to a "references.txt" file (or something) for each community code though. That would be more accessible than the current method of parsing the docstring (since no one knows exactly how this parsing is done).

rieder:
That would have other benefits: we could use a documentation standard (e.g. bibfiles, or something more recent/advanced), and just parse these for output and/or provide a bibfile that AMUSErs can just include in their paper directly. Perhaps similar to the "reproducible paper" method from Akhlaghi.

spzwart:
Yes, enter the entire bibtex blurb is probably best.
What about adding a file called: interface.bib
or -codename.bib-
These could be included directly in any latex document.

rieder:
Yes, "interface.bib" is probably the easiest way to do this. One caveat is that we would have to process the file as well, which adds some dependencies to AMUSE... Not sure how to do this yet (I don't want TeXLive as a dependency for AMUSE ;) ).

rieder:
@ipelupessy If I remember correctly you created an issue about something related to this a year ago or so... About some documentation format? I forgot exactly what it was and can't find it...

ipelupessy:
hmmm, does not directly ring a bell...I think adding the ADS codes is fine and enough for the moment...

rieder:
I was thinking of issue #232 (which is something different than this after all).

ipelupessy:
is probably ok to merge?

rieder:
This one isn't "done", but it's an improvement already so I'm merging it.

PR #471, opened 2019-09-22
Amuse 13

rieder:
This PR tracks work for AMUSE 13 that removes Python 2 compatibility and makes AMUSE a native Python 3 package.
After merging this PR, AMUSE will no longer work with Python 2.


rieder:
- Need to fully include fixes to change hermite0 -> hermite (seems some commits were not included)
- Need to check everything works as expected (I successfully tested installing framework & run Hermite)

ipelupessy:
I think we should make a python3 branch in the main repository..also I think al the files need to be hand checked, because the auto conversion generates a lot of ugly and inefficient code constructs...

ipelupessy:
I guess we could use the review option to keep track of the work done..but it would maybe help to cut the commits in manageable groups (e.g. tests would not need the same attention, interface py files are generally also ok, but ofcourse the core things should be checked carefully...)

rieder:
> I think we should make a python3 branch in the main repository..also I think al the files need to be hand checked, because the auto conversion generates a lot of ugly and inefficient code constructs...

Sure, let's do it in the main repository. Consider this one an early test. This code is nearly the same as the code of AMUSE 12 when installed in Python 3, which certainly isn't what it definitively should be.

rieder:
Of course for this we should follow the Zen of Python (https://www.python.org/dev/peps/pep-0020/), but also "done is better than perfect".

rieder:
Anyway, I'm in Leiden Tue-Fri to work on this.

ipelupessy:
maybe we should do the following:
- python 3 branch on the main repository with no changes
- breakdown the PR in multiple PRs of branches from your repo to the main python3 branch
- then sign off on files in the PR  in code review until all files are done...

rieder:
Sounds good!

ipelupessy:
so I have created the branch...

ipelupessy:
btw, I don't think py23 converts the doc string tests??

rieder:
... I'm not sure? Where are those located?

ipelupessy:
in the code;-) we will see, its easy to fix if we run into them...

rieder:
Obsolete branch, closing.

PR #473, opened 2019-09-24
Update lab.py to not fail on non-installed modules

rieder:


rieder:
@ipelupessy is this OK?

PR #474, opened 2019-09-24
2to3 conversion of rfi/

rieder:
Initial automatic conversion. Need to check for ugly constructions etc., but this code is essentially what AMUSE-12 currently looks like under Python 3.

### DIFF
--- a/src/amuse/rfi/async_request.py
+++ b/src/amuse/rfi/async_request.py
@@ -1,10 +1,10 @@
 import select
 import operator
 
-import channel
+from . import channel
 
 class AbstractASyncRequest(object):
-    def __nonzero__(self):
+    def __bool__(self):
         return not self.is_finished
 
     def waitone(self):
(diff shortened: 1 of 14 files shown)


rieder:
> just to be sure - this is what setup.py generates in the build directory with python setup.py build?

Yes, the files are identical.

ipelupessy:
I think the changes I have made are safe...

ipelupessy:
I am scratching my head currently on the iteritems -> items and replacement with keys (there are AMUSE classes which implement these...) So I am not sure the auto 2-->3 changes work in these cases (the heuristics for the changes may get it wrong?)

rieder:
well, it should /work/ (because it's tested with AMUSE 12), but not sure it's how we want it to be.  Let's discuss tomorrow?

ipelupessy:
I think we can just be optimistic, because if it fails it gives quite clear error...

PR #477, opened 2019-09-24
Python3 datamodel

rieder:


src/amuse/datamodel/base.py
```diff
@@ -616,15 +616,15 @@ def __setstate__(self, data):
...
+        for name, value in self._attributes.items():
             lines.append("{0}: {1}".format(name, value))
         return '\n'.join(lines)
     
     def _copy_for_collection(self, newcollection):
         return CollectionAttributes(self._attributes)
         
     def iteritems(self):
-        return self._attributes.iteritems()
+        return iter(self._attributes.items())
```
ipelupessy: this one I dont understand...

ipelupessy: also it looks like the name of the method is dangerous..on the otherhand, if this goes wrong it will generate an easy debugged error..

src/amuse/datamodel/incode_storage.py
```diff
@@ -662,7 +662,7 @@ def apply_on_all(self, particles, *list_arguments, **keyword_arguments):
             
         if not self.get_number_of_particles_in_set_method is None:
             number_of_particles_in_set = self.get_number_of_particles_in_set_method(*query_identifiers)
-            indices = self.method(range(number_of_particles_in_set))
+            indices = self.method(list(range(number_of_particles_in_set)))
```
ipelupessy: not sure about this list, keeping it is safe..

src/amuse/datamodel/particles.py
```diff
@@ -281,33 +281,33 @@ def to_string(self, attributes_to_show = None, split_at = 20):
                 if isinstance(quantity, LinkedArray):
                     values_to_show = list(map(format_str11,quantity[:split_at].to_print_list()))
                     values_to_show.append(format_str11('...'))
-                    values_to_show.extend(map(format_str11,quantity[-split_at:].to_print_list()))
+                    values_to_show.extend(list(map(format_str11,quantity[-split_at:].to_print_list())))
```
ipelupessy: extend for a list can take iterator, so list here and below is not necessary

src/amuse/datamodel/base.py
rieder: Related to #489 in that it is caused by 2to3 possibly being too careful

PR #478, opened 2019-09-24
2to3 conversion of io/

rieder:


rieder:
A general note though: the code as converted with 2to3 is the same as what is used currently in Python3 Amuse (v12). In that sense the changes are already "safe".

ipelupessy:
That is true, however the auto convert generates a lot of unnecessary list - I think we should try to fix it right away...

PR #479, opened 2019-09-24
2to3 conversion of src/amuse/support/

rieder:


src/amuse/support/core.py
```diff
@@ -202,10 +202,10 @@ def string(self):
...
     def isstring(self, x):
-        return isinstance(x,types.StringType)
+        return isinstance(x,bytes)
         
     def isnumber(self, x):
-        return isinstance(x,types.IntType) or isinstance(x,types.FloatType) 
+        return isinstance(x,int) or isinstance(x,float) 
         
         
 class OrderedDictionary(object):
```
ipelupessy: this one is tricky because the py23 converter has tried to maintain effectively python 2 behaviour, while the class mimicks a dict, so probably should use python3 dict conventions...

in any case nowadays there is a build-in ordered dict, unfortunately it behaves not exactly the same

src/amuse/support/state.py
```diff
@@ -176,7 +176,7 @@ def remove_transition(self, from_name, to_name):
             to_state.remove_to_transition(from_state)
             
     def iter_states(self):
-        return iter(self.states.values())
+        return self.states.values()
```
ipelupessy: this should be ok

src/amuse/support/core.py
rieder: #489

PR #480, opened 2019-09-24
2to3 community/

rieder:
Ignore all python files from src/ directories though (if these existed)


ipelupessy:
are there any src python files? I am not sure if those are converted...

src/amuse/community/athena/downloadhttp.py
```diff
@@ -4,7 +4,7 @@
 import os
```
ipelupessy: there is a seperate issue for the dowload scripts..

src/amuse/community/galactics/gas_interface.py
```diff
@@ -764,11 +764,11 @@ def update_particle_set(self):
         number_of_updated_particles,number_of_updated_gas_particles = self.get_number_of_particles_updated()
         if number_of_updated_particles:
             self.particles._private.attribute_storage._add_indices(
-                range(number_of_updated_gas_particles,number_of_updated_particles)
+                list(range(number_of_updated_gas_particles,number_of_updated_particles))
```
ipelupessy: this is a recurring pattern: functions should be able to handle iterators, and then the list here can be removed...

  we should generate an issue for this

ipelupessy:
> are there any src python files? I am not sure if those are converted...

in any case: there is a question whether we should convert a community python 2 code..you can envision interfacing from python3 to a python2 code!!

src/amuse/community/interface/se.py
```diff
@@ -922,7 +922,7 @@ def get_density_profile(self, indices_of_the_stars, number_of_zones = None):
         indices_of_the_stars = self._check_number_of_indices(indices_of_the_stars, action_string = "Querying density profiles")
         if number_of_zones is None:
             number_of_zones = self.get_number_of_zones(indices_of_the_stars)
-        return self.get_density_at_zone([indices_of_the_stars]*number_of_zones, range(number_of_zones) | units.none)
+        return self.get_density_at_zone([indices_of_the_stars]*number_of_zones, list(range(number_of_zones)) | units.none)
```
ipelupessy: what does a iterator + unit do? (generate the list for the quantity?)

src/amuse/community/sakura/ggf/old_test_sakura.py
```diff
@@ -17,137 +17,137 @@ class TestSakuraInterface(TestWithMPI):
     def test01(self):
         if MODULES_MISSING:
             self.skip("Failed to import a module required for Sakura")
-        print "Test SakuraInterface initialization"
+        print("Test SakuraInterface initialization")
         instance = SakuraInterface()
-        self.assertEquals(0, instance.initialize_code())
-        self.assertEquals(0, instance.commit_parameters())
-        self.assertEquals(0, instance.cleanup_code())
+        self.assertEqual(0, instance.initialize_code())
```
ipelupessy: I am not sure  about the replacement of assertEquals with assertEqual - should check TestWithMPI

ipelupessy: I am not sure it matters here (see filename), but I have seen it elsewhere

rieder:
> > are there any src python files? I am not sure if those are converted...
> 
> in any case: there is a question whether we should convert a community python 2 code..you can envision interfacing from python3 to a python2 code!!

I found them in Ramses. But I think they're used as analysis tools only. I don't want to touch them in principle, we should maybe get an updated upstream version though.

src/amuse/community/sakura/ggf/old_test_sakura.py
rieder: We may want to use this opportunity to do some cleanup... Also for ph4.

rieder:
Don't replace ints with longs - longs no longer exist in Python 3 (https://www.tutorialspoint.com/python3/python_numbers.htm)

ipelupessy:
the line before defines long as a vector...!!!

rieder:
> the line before defines long as a vector...!!!

Ah, I see. That is just evil... We should rename it to something more sensible than a reserved keyword.

rieder:
But well spotted!

src/amuse/community/athena/downloadhttp.py
rieder: yes - will address that in a separate PR...

src/amuse/community/galactics/gas_interface.py
rieder: yep... Not sure how to formulate it yet...

rieder:
longv is much better. thanks.

src/amuse/community/sakura/ggf/old_test_sakura.py
rieder: > I am not sure about the replacement of assertEquals with assertEqual - should check TestWithMPI

This is a "preferred" naming change in Py3, but both should work in Py2 and Py3 iirc.

rieder: https://docs.python.org/3/library/2to3.html?highlight=assertequal#2to3fixer-asserts

ipelupessy: ok, it does not matter, in test/amusetest.py:
```
    assertEqual = failUnlessEqual
    assertEquals = failUnlessEqual
```

src/amuse/community/interface/se.py
rieder: list + unit gives a VectorQuantity. Here, the list is very important:
```
In [1]: from amuse.units import units                                                                                                                     

In [2]: a = list(range(4)) | units.m                                                                                                                      

In [3]: a                                                                                                                                                 
Out[3]: quantity<[0, 1, 2, 3] m>

In [4]: b = range(4) | units.m                                                                                                                            

In [5]: b                                                                                                                                                 
Out[5]: quantity<range(0, 4) m>
```

rieder: So b is a very weird ScalarQuantity in this example

src/amuse/community/galactics/gas_interface.py
rieder: Issue #488

src/amuse/community/interface/se.py
rieder: Issue #484 and #488

PR #481, opened 2019-09-24
2to3 couple/

rieder:


src/amuse/couple/parallel_stellar_evolution.py
```diff
@@ -127,7 +127,7 @@ def run(self):
     
     def get_result(self):
         if self.caught_exception:
-            raise self.result[1], None, self.result[2]
+            raise self.result[1].with_traceback(self.result[2])
```
ipelupessy: wondering about this construct

rieder: Here's a bit of documentation about that change: https://docs.python.org/3/library/2to3.html?highlight=raise#2to3fixer-raise

PR #487, opened 2019-09-27
Implement opportunistic deadlock avoidance for worker crashes.

merijn:
Amuse has a tendency to deadlock when the remote worker crashes, due to the
worker(s) never calling MPI_Finalize or MPI_Abort. This change installs (only
on POSIX platforms and only on thread-safe MPI initialisations) a signal
handler that traps some of the more common "crash" signals (SIGABRT, SIGBUS,
SIGILL, SIGINT, SIGQUIT, SIGSEGV, and SIGTERM) and attempts to call MPI_Abort()
before dying.

This is not sufficient to avoid deadlock, as the MPI_Abort() never notifies
the MPI_Comm_spawn parent that the run is aborted. The parent (i.e., amuse) is
generally blocked in an MPI_Wait, waiting for a return message from a function
call, due to the way mpi4py is implemented amuse becomes (nearly)
uninterruptible by keyboard interrupts (SIGINT), nor does it detect the
MPI_Abort in the worker.

To work around this, the worker sends a non-blocking all 0 message to the
parent (waking it up from the MPI_Wait), this message fails the call and
function id checks on lines 1149 and 1152 of amuse/rfi/channel.py, resulting in
a call to MPI_Comm_disconnect in the parent. The worker also calls
MPI_Comm_disonnect and only then calls MPI_Abort to end the other worker tasks.

This way both the workers and clients abort when an error/crash happens. This
doesn't always work (the disconnect could block, some other MPI error could
happen while trying to send/abort, etc.), but since the current behaviour is to
just deadlock, we're not off any worse than before and when it does work things
are much more pleasant to test with.

### DIFF
--- a/src/amuse/rfi/tools/create_c.py
+++ b/src/amuse/rfi/tools/create_c.py
@@ -41,6 +41,10 @@ HEADER_CODE_STRING = """
 	#include <netinet/tcp.h>
   #include <arpa/inet.h>
 #endif
+#if _POSIX_VERSION >= 1
+#define _POSIX_C_SOURCE 1
+	#include <signal.h>
+#endif
 """
 
 CONSTANTS_AND_GLOBAL_VARIABLES_STRING = """
@@ -96,7 +100,6 @@ static char * * strings_out;
 static char * characters_in = 0;
 static char * characters_out = 0;
 """
-
 
 POLLING_FUNCTIONS_STRING = """
 static int polling_interval = 0;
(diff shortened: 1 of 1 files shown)


ipelupessy:
can this be implemented for fortran codes too?

ipelupessy:
I mean eventually? (no need to do this right now...)

merijn:
I don't see a reason why it couldn't be

ipelupessy:
if it works, is there a exception generated on the amuse side?

merijn:
Yeah, the checks I mentioned in channel.py end up throwing an exception.

ipelupessy:
I will put out an issue for fortran...

ipelupessy:
One thing I was wondering: can something like this also be made for workers that keep on spinning after python side has shutdown?

ipelupessy:
just an additional question:
If I don't overlook something, the signal thing does nothing when sockets channel is used? also: is the code correct if started if compiled with MPI (so NOMPI off), but with the sockets channel?

PR #490, opened 2019-09-27
2to3 test/

rieder:


ipelupessy:
is it possible to break this down in:
- assert replacements and prints
- and other stuff
and maybe ignore/ disable list around .values()..I have got the feeling this should work ok automatically

(there might not be much diff left after the this)

ipelupessy:
(this for the actual tests, not for the code in src/amuse/test)

rieder:
> (this for the actual tests, not for the code in src/amuse/test)

I included those later...

rieder:
> is it possible to break this down in:
> 
> * assert replacements and prints
> * and other stuff
>   and maybe ignore/ disable list around .values()..I have got the feeling this should work ok automatically
> 
> (there might not be much diff left after the this)

Sure, but I'll wait for Travis to finish checking this one first.

ipelupessy:
there are failures, osx reports as passed, but also has errors...
nothing overly serious though

rieder:
I created a separate PR for print+asserts changes, need to make a third one which re-applies the other changes onto that PR... This is a bit tricky, will do it later.

ipelupessy:
closing this one...

PR #495, opened 2019-09-30
non print and assertion 2to3 conversions for test directory

ipelupessy:


support/misc.py
```diff
@@ -3,13 +3,10 @@
...
 import fnmatch
 
-if sys.hexversion > 0x03000000:
-    from os import walk as py_walk
-    def walk(top, callback, args):
-        for root, dirs, files in py_walk(top):
-            callback(args, root, files)
-else:
-    from os.path import walk
+from os import walk as py_walk
```
ipelupessy: how did this commit end up here?

Issue #412, opened 2019-05-07
Amuse package requirements


rieder:
Currently, amuse-framework has all the AMUSE prerequisites, though it only needs Python (and python related things) itself. 
Maybe we can be more flexible about this in the future - e.g. amuse-framework has very few requirements, C++-based codes require a C++-compiler, Fortran codes a Fortran compiler, etc.
This would probably require splitting up the configure script (or at least its output).

ipelupessy:
I find this tricky, since you can also use the installed amuse framework as a basis to develop....
(though some libraries might be dropped)

rieder:
How about splitting amuse-framework one more time, into
- amuse-core (only the Python bits that need no compiling or configuring at all)
- amuse-libraries (the other bits)

ipelupessy:
yes, I like the idea. I would retain amuse-framework as high level package with these two. Maybe a better name for amuse-libraries??  (don't have a picture of the exact split..)

rieder:
I'll have a go at this.

ipelupessy:
ok, be careful though..I think the config is required in unexpected places...

Issue #424, opened 2019-05-15
Uninstalling a package can remove other files


rieder:
**Describe the bug**
Uninstalling an AMUSE package can (in some cases) lead to essential files being deleted as well - e.g. removing amuse-ph4 would delete all the amuse-framework stuff (but not uninstalling the package). The package still lists as being installed, but the module is no longer present.

**To Reproduce**
Will try to figure this out later, I just encountered this a few times now.

**Expected behavior**
Only the files belonging to the specific package should be deleted

**Environment (please complete the following information):**
 - OS and version: macOS Mojave
 - Python: 2.7

rieder:
Seems related to an outdated MANIFEST.in

rieder:
Seems this may not be completely solved

PR #498, opened 2019-10-01
Add "amuse-core" package which installs python-only bits of AMUSE

rieder:
An attempt at fixing #412, splits of amuse-core package from amuse-framework.

Work in progress, overlapping files need to be removed from amuse-framework to prevent conflicts.


rieder:
This will be much easier once we only support one Python version, so this is for after 13.0 is released.

ipelupessy:
amuse-framework with dependency on amuse-core?

rieder:
yes

ipelupessy:
ok, don't worry about the commits, I am adding something but merged master first..

ipelupessy:
there is now a seperate core install mode which should handle core installs properly (I don't think it actually changes anything, but now we can add things if we want)

ipelupessy:
I have tried this together with #501 (it needs something to prevent the config error), and it installs and allows you to do *something* (importing units)

ipelupessy:
I have tried this together with #501 (it needs something to prevent the config error), and it installs and allows you to do *something* (importing units)..its fairly annoying though that if you make amuse-core a requirement for amuse-framework a purely local install does not work..

ipelupessy:
ok, this is fixable with --find-links option to pip, next:
(development install)
```
pip install -e . 
```
does not work casue it cannot find the config
(I am trying out an amuse-framework edit which I didn;t commit yet)

ipelupessy:
I have tried to make changes to amuse-framework, it has the above config problem (which is fixable), but more seriously, the uninstall does not work properly (pip uninstall of amuse-framework removes core components) ---this indicates something seriously wrong with the way we implement this.l..

rieder:
Hmm, yes, that would be a problem... Wondering if this is the same as what I found in #424? I was never really sure about that one...

ipelupessy:
How was #424 resolved? I don't know how the removal of a package works..

rieder:
It wasn't resolved (reopened now), but after updating manifest I couldn't reproduce it... Also didn't test it all that much.

rieder:
Relevant documentation for packaging: https://packaging.python.org/guides/packaging-namespace-packages/
I don't think we're doing it right currently...

ipelupessy:
hmm..on the other hand, the problem we encunter doesn't look like anything that is described as failure modes in the link...

ipelupessy:
I think this is currently low priority, maybe defer? @rieder

ipelupessy:
close due to low priority

PR #499, opened 2019-10-01
Allow using AMUSE without configuration

rieder:
An exception here is perhaps too harsh...
A core-only installation will not need (and in fact should not have) a config file.
We should think of a better way to handle this.

### DIFF
--- a/src/amuse/__init__.py
+++ b/src/amuse/__init__.py
@@ -38,4 +38,4 @@ numpy_fix()
 try:
     from . import config
 except:
-    raise ImportError("amuse is not configured correctly")
+    print("amuse is not configured correctly")  


ipelupessy:
well, the problem with this is silent failures vs annoying messages..it should fail when a confiuration is expected and not when no configuration is expected to be present...

rieder:
yes, exactly. 
For an "amuse-core" package, all is fine if no config file is found, so there should not even be an error message. But for amuse-framework (i.e. the libraries needed for community), the config is essential.

ipelupessy:
if you move the check (raising an exception) to amuse-framework (a seperate file and a import with try/except in __init__.py would work) that would work I think...

rieder:
Yes, I guess that's an option. It like a bit of a hack though... I'll think about this for a bit (other things to do first).

ipelupessy:
check on presence of config.mk and make setup_codes.py so that it always generates a config.mk even if configure fails?

ipelupessy:
an alternative is #501 - I may still like the simpler approach of adding a file

rieder:
I think it should be as follows
- amuse-core (python only) needs & installs no config, and works without it (also no error message)
- amuse-framework (amuse-core + libraries + config) installs config and check for config
- if the config is missing or "wrong" after installing amuse-framework, it displays an error message that doesn't stop AMUSE from working (i.e. #501).

So I think #501 in combination with a separate file (installed by amuse-framework) imported with try/except would solve this.

ipelupessy:
#501 by itself works, it does generate an error if the config dummy that is generated is actually used - but that is fine, because there may be places in the code where it is called while using core (if we encounter these, this then indicates someting should be refactored (because the config is actually not needed) or moved to amuse-framework)

rieder:
OK, fair point.
I think #501 is a much better solution than this one, closing this PR.

PR #501, opened 2019-10-03
allow amuse to run w/o config (until it is needed)

ipelupessy:
alternative to #499


ipelupessy:
mostly harmless, so merged

PR #515, opened 2019-10-17
Allow usage of pc

cgroeneveld:
Allow usage of pc instead of only parsec, so to be consistent with Gpc/Mpc/kpc


ipelupessy:
seems ok to me..

PR #521, opened 2019-10-30
make mpi optional for amuse-framework

ipelupessy:
this needs to be tested


ipelupessy:
it would be nice to have amusifier report what MPI functionality the installation has...

ipelupessy:
now only amuse-framework has the option of being pip installed w/o mpi4py...

ipelupessy:
ok, I don't think we need codecov report here!!

rieder:
No that seems a bit much...

ipelupessy:
I think this PR might actually work...

ipelupessy:
but it is nice that codecov works..I looked at the report and it reports detailed info, can be very handy

rieder:
yes I think so too. I've now disabled the comments, but it'll still work.

PR #551, opened 2019-12-11
fix gravitational dynamics state for set_mass etc

ipelupessy:


ipelupessy:
needs tested against all gd codes

rieder:
This solves the problem with BHTree for me.

rieder:
perhaps also set_radius should be included?

ipelupessy:
have you tried other codes? though I think if it breaks any of the other codes, these are at fault...

rieder:
> have you tried other codes? though I think if it breaks any of the other codes, these are at fault...

ph4 seems fine, I have not yet tried others (will do so later). I agree this should work and if it doesn't the other codes are at fault.

ipelupessy:
merge?

rieder:
I'll run the tests for the other gravity codes, if they run OK I'll merge.

rieder:
All tests I ran passed - merging.

PR #554, opened 2019-12-18
fi minor change, move abs acc to seperate array 2

ipelupessy:
attempt 2 (rebased on recent master


MJCWilhelm:
A run of fi including gravity and hydrodynamics (no black holes) gave the exact same results between this version and the standard version.

ipelupessy:
ok, thanks..my tests where also ok.. I think its good to merge..blackhole stuff is rarely used anyway (also patch works the same way)

PR #556, opened 2019-12-18
added version of bridge that includes velocities of individual stars …

webbjj:
Note that the changes to bridge itself are minor. It just needs a use_velocities flag and self.vx,self.vy,self.vz need to be passed in get_gravity_at_point. I don't think they are needed in other functions. So it may not be necessary to have a separate file, and bridgev could be folded into bridge.

### DIFF
--- a/src/amuse/couple/bridgev.py
+++ b/src/amuse/couple/bridgev.py
@@ -1,0 +1,735 @@
+"""
+  bridge-like integrator for amuse
+
+  the bridge class provides a bridge like coupling between different
+  gravitational integrators. In this way a system composed of multiple
+  components can be evolved taking account of the self gravity of the whole
+  system self consistently, while choosing the most appropiate integrator
+  for the self-gravity of the component systems. This is mainly useful for
+  systems  consist of two or more components that are either well separated
+  spatially or have different scales (otherwise using a single integrator is
+  more efficient)
+
+  The main idea is that systems experience each others gravity through
+  periodic velocty kicks with ordinary evolution in  between - the evolution
+  is thus described by an alternation of drift (D) and  kick (K) operators,
+  here chosen as:
+
+       K(1/2 dt) D(dt) K(1/2 dt)
+
+  K(dt) denotes a kick of the velocities over a timestep dt, while D(dt)
+  denotes  a drift, meaning secular evolution using self gravity of the
+  system, over dt.
+
+  implementation notes:
+
+  In order to use bridge the component systems should be initialized as usual,
+  then a bridge systems is initialized, after which one or more systems are
+  added:
+
+  from amuse.ext.bridge import bridge
+
+  bridgesys=bridge(verbose=False)
+  bridgesys.add_system(galaxy, (cluster,), False)
+  bridgesys.add_system(cluster, (galaxy,), True )
+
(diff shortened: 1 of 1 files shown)


rieder:
I would definitely fold it into bridge.py, and make it optional to pass the velocities. It would also make the changes easier to spot for us :).

I think the use_velocities flag should be passed when adding a "kicker" system, individually for each of these. Not all system may support get_gravity_at_point with velocities.

ipelupessy:
great! I would also put the changes in new (derived) classes in this case (and a more descriptive name than bridgev) in bridge.py. Can you also add an example of its use and some tests wouldalso  be nice? 

also next time: i suggest you make a feature branch in your fork and a PR against this (this makes it easier for yourself to work on multiple things at the same time;-)...)

webbjj:
> I would definitely fold it into bridge.py, and make it optional to pass the velocities. It would also make the changes easier to spot for us :).
> 
> I think the use_velocities flag should be passed when adding a "kicker" system, individually for each of these. Not all system may support get_gravity_at_point with velocities.

Where is the check to see if the system is a "kicker" or not? Its not entirely clear to me where this is in bridge.

webbjj:
> great! I would also put the changes in new (derived) classes in this case (and a more descriptive name than bridgev) in bridge.py. Can you also add an example of its use and some tests wouldalso be nice?
> 
> also next time: i suggest you make a feature branch in your fork and a PR against this (this makes it easier for yourself to work on multiple things at the same time;-)...)

Thanks for the suggestion. I am still a bit of a git rookie.

I was also hoping you can elaborate on your initial suggestion to put the changes in new (derived)  classes. Are you saying make a class Bridgev (with better name) as opposed to adding the use_velocities option to class Bridge? Similarly, add a class GravityCodeInFieldv instead of adding to GravityCodeInField?

ipelupessy:
ok, I checked more carefully now: you should indeed make a new:
```
class BridgeWithVelocityDependendForces(Bridge):
..etc
```
(or some other descriptive name)..the option use_velocity still has to be present because as it is now you can enable these on a per system basis. 

eventually this could be folded back in the Bridge, but I think it is nice to have a new class for now since that makes it clearer what the differences are.

already mentioned this, but tests and a way to validate would also be nice;-)

rieder:
I made some changes in a PR to this branch, creating a separate VelocityDependentBridge class (which mostly just inherits from Bridge) and merging the other changes into bridge.py.

rieder:
I'm not sure if I like the re-defining of "get_gravity_at_point" with velocities.

Perhaps we should have a new function "get_force_at_point" for this purpose - after all, the velocity dependent forces are different from gravity...

ipelupessy:
one thing I don't fully like: the method to get the acceleration in the velocity kick case is still called get_gravity; maybe it should be renamed to get_acceleration?

ipelupessy:
ok, I see you had the same concern ;-)

rieder:
@ipelupessy should we perhaps just move to get_acceleration_at_point altogether, leaving get_gravity_at_point as an alias for this in case it is different?

rieder:
Reminder to self that this is close to being mergeable. Just needs a name change for some functions.

ipelupessy:
keep open for now

ipelupessy:
keep

PR #557, opened 2019-12-18
work on update secular multiple

ipelupessy:
not for merging yet


ipelupessy:
it is still a good idea; best done by someone who needs the updated version though

ipelupessy:
keep

ipelupessy:
so it needs a bit of work to reconsolidate the interface of amuse with the interface.cpp of secular multiple

rieder:
The original developer of Secular Multiple has left the field now, I'm not sure if development continues elsewhere (I think it doesn't for now).
It would be good to test/finish this, since at least we then have the latest version included. It seems to be a lot cleaner than the original version.

rieder:
As for the interface conflicts - best to use the version from this PR, we can clean it up again later.

src/amuse/community/secularmultiple/Makefile
```diff
@@ -10,31 +10,46 @@ MPICC ?= mpicc
...
 	$(CXX) $(CXXFLAGS) -c -o $@ $<
 
 .c.o: $<
 	$(CC) $(CXXFLAGS) -c -o $@ $<
+
+src/interface.cpp:
+	make -C . download
+
+download:
+	git clone -b master https://github.com/hamers/secularmultiple.git src
```
rieder, line 55: We should download a specific version rather than just the latest (even though it's no longer updated here).

PR #558, opened 2019-12-19
Corrected bug when committing zero-length particle sets.

stevemcmillan:


rieder:
Looks fine - merging.

PR #569, opened 2020-01-23
remove old kickoff presentation pdfs

ipelupessy:
I guess there is no real reason to have these in the repository..if we want to have them its maybe better to move them somewhere else?


rieder:
Yeah I agree.

ipelupessy:
its kind of nice to have it somewhere for historical reasons (I recently showed some of the old slides to illustrate where we came from..)..can we host files on the website?

ipelupessy:
otherwise the sandbox?

rieder:
Either the website or the sandbox would be a better place. I think sandbox is the best place - the website should preferably include more up to date things.

ipelupessy:
done

PR #570, opened 2020-01-23
remove omuse tutorial (added to OMUSE repo)

ipelupessy:


ipelupessy:
the OMUSE binary build need to be fixed as a result, but the build scripts need update anyway...

PR #581, opened 2020-02-04
Refactoring kira.py to follow the AMUSE coding standard more closely

rieder:


rieder:
This PR is only syntax related, no actual fixes.

Issue #559, opened 2019-12-19
Make amuse.lab feature-complete for the textbook examples


rieder:
amuse.lab should have all the basics available for importing (generally speaking: all the things that are required for the examples in the book)

PR #582, opened 2020-02-04
Add alternative interface names, update lab.py

### LINKED ISSUE #560: Make community codes available under a generalised name
- Community codes should be available under their current name (with not necessarily predictable capitalisation) as well as a spelling with only the first letter capitalised (e.g. Bhtree, Simplex, Ph4).

rieder:
Fixes #560 (additional fixes are needed for non-lab.py codes)
Work in progress for #559


rieder:
Tested, works for me.

PR #586, opened 2020-02-11
fix syntax of galaxia interface

rieder:
Fix numerous code style problems (indentation, imports, spaces). No functional change.

### DIFF
--- a/src/amuse/community/galaxia/interface.py
+++ b/src/amuse/community/galaxia/interface.py
@@ -1,1339 +1,1274 @@
 """
 Interface Definition for the galaxy models
 """
-import numpy
-import os
-
-from amuse.units import nbody_system
-from amuse.units.quantities import VectorQuantity
-from amuse.community import *
+from amuse.units import units
+from amuse.community import CodeInterface
 from amuse.community.interface.common import CommonCodeInterface, CommonCode
-from amuse.community.interface.gd import GravityFieldInterface, GravityFieldCode, GravitationalDynamics
+from amuse.community.interface.gd import GravityFieldCode
 
 from amuse.rfi.core import legacy_function
 from amuse.rfi.core import LegacyFunctionSpecification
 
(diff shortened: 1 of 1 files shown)


ipelupessy:
I think the new version has a problem with the indentation of multiline line function arguments, ie:
```
   function.addParameter('x', dtype='float64', direction=function.IN
                            , description="x position",unit=units.kpc)
```
is transformed to:
```
   function.addParameter('x', dtype='float64', direction=function.IN
, description="x position",unit=units.kpc)
```
instead of
```
   function.addParameter('x', dtype='float64', direction=function.IN,
                                            description="x position",unit=units.kpc)
```

ipelupessy:
please also remember that these kind of changes basically reset the file history (for e.g. blame)...

ipelupessy:
ok, discard my earlier comment on the indent

ipelupessy:
it was just the github viewer..

ipelupessy:
or to be more precise: me misreading it...

rieder:
> please also remember that these kind of changes basically reset the file history (for e.g. blame)...

Yes, this shouldn't be done too often, but sometimes it's necessary :).
The file history can still be retrieved, and it's exactly the reason why this change is specifically only targeting syntax changes.

ipelupessy:
yea, in this case probably ok, but you can verify for yourself that the
[blame](https://github.com/amusecode/amuse/blame/e192c718f9de9b98d5113c3eeb46e3e301cbe38f/src/amuse/community/galaxia/interface.py) is not exactly informative now;-)

rieder:
Not in master, but you could easily go back to an earlier version (e.g. right before this merge, https://github.com/amusecode/amuse/tree/d597c0f0935b5afdd4d5a45b1b90619436b33e56) and check changes there.
Mucking up git blame a bit can't in itself be a reason not to make changes ;-).

rieder:
Not that blame on this interface would be that much more informative then...

Issue #587, opened 2020-02-12
Interface compilation ambiguities


MJCWilhelm:
With a recent update, the way interfaces are compiled has changed. Before, it was done through 'make clean/make' in the directory of an interface inside src/amuse/community, but that method no longer works (giving an error in build.py, where it can't find the AMUSE module). Currently, it is done through 'make {code}.code' in the main AMUSE directory. 

This results in an ambiguity when making new interfaces. Specifically, the makefile generated by the build.py script in dir mode uses the amusifier executable to compile an interface, instead of build.py. This does still work with the old make method, but not with the new method. In all new interface makefiles, the line 'CODE_GENERATOR = $(AMUSE_DIR)/build.py' has been added, which means build.py will be used for compilation instead of amusifier. This does work with the new method, but not the old. 

It would be most straightforward to change the generated makefile to include the new definition and state in the tutorial that the new interface should be generated in the community folder, but I can also see that one would want to develop an interface in another place than the amuse directory. In either case, the interface tutorials should be edited to reflect this ambiguity.

Any thoughts?

ipelupessy:
hmm, the make inside the code directory should still work (i do this all the time)..specifically it should always be possible to make a makefile which works in any location (like you say) with a make in the dir.
you may need to set some environment variables though...can you give an example of a code that does not work? (it may also be the makefile) and the error? (and how is amuse installed: I assume a straight clone of the repo with ./configure and make framework done?)

build.py should be the same as amusifier (except that amusifier lifes in the bin directory, you should set a PATH for it to work)..build.py is often found by some relative AMUSE_DIR specification (this is unreliable) in the makefiles

MJCWilhelm:
I tried this on both sse and ph4, the error I get is:

```
/data1/wilhelm/py_envs/amuse_dev/amuse//build.py --type=c interface.py ph4Interface -o worker_code.cc
Traceback (most recent call last):
  File "/data1/wilhelm/py_envs/amuse_dev/amuse//build.py", line 7, in <module>
    from amuse.rfi import gencode
ModuleNotFoundError: No module named 'amuse'
make: *** [Makefile:61: worker_code.cc] Error 1
```
The installation is indeed done through ./configure and make (but not 'make framework' I think?). 

[...]

PR #588, opened 2020-02-13
make all makefiles use amusifier

ipelupessy:
fixes some inconsistent build issues


ipelupessy:
I need to know if I got all references and whether packages still work..

rieder:
Codes are building fine for me - both with 'make X.code' and in the code directory itself (using the "pip install -e ." method to install AMUSE).

ipelupessy:
oops, they should have the same (they have = instead of =? so missed them)

rieder:
> oops, they should have the same (they have = instead of =? so missed them)

Ok, I changed them.

ipelupessy:
thanks!

rieder:
I'm happy for this to be merged.

ipelupessy:
waiting whether @MJCWilhelm has any comments (should solve his problems..)

MJCWilhelm:
With this change the './configure make' only works partially; only asterisk, gadget2, galactics, higpus and mameclot are successfully built. The others I can't build through 'make code.code', although some can be built in their community directories (tested this with bse, which didn't build, and ph4 and fi, which did). Also, building gadget2 (one of the ones that did build at first) in its directory did work.

I still suspect that it has something to do with how amuse runs on the strw pc's (using 'mpiexec $AMUSE_DIR/amuse.sh' instead of just 'python'). My errors are that amuse can't be imported, which is exactly what happens when you try to import amuse in a normal python session on these pc's.

ipelupessy:
ok, did you follow either (which?) of the procedures I lined out in #587? (setting environment variables or using virtualenv & pip install -e . - the environment variables are needed because otherwise python will not be able to import and the amusifier script needs to be in the PATH, if you use the pip development install, pip takes care of this (while linking to the development dir instead installing the files))

MJCWilhelm:
I did use a virtual environment, but built amuse using the old './configure / make' method, since the pip version doesn't work on the strw machines. Amuse does build, but it gives errors when trying to run (even freezing when trying 'python script.py'). I did see that amusifier was added the venv's bin, which it wasn't before.

ipelupessy:
ok, in that case you have to set the environment variables!

MJCWilhelm:
I set the environment variables you mentioned and then both the main folder and code folder makefiles work. 

With this change, rebuilding codes does become a bit annoying when using the old installation method (having to set two extra environment variables). Is there a good way to append to them automatically? Perhaps the makefile can check whether they are in the (PYTHON)PATH?

rieder:
I wouldn't recommend the old method anymore. I'd much prefer fixing the new method for where it doesn't work currently to messing with environment variables.

ipelupessy:
ok, great, I take it it is working then!

what I usually do is putting the environment variables (or virtualenv sourcing and other stuff (such as selecting the mpi flavour)) in a file named setdev in the root of a project folder..its always a matter of sourcing a single file then...note that from the viewpoint of amuse whichever method you use for setting up your environment should work (whether you do it fully manually, or partially or wholly managed by virtualenv or other tools..)..if something seemed to work before w/o configuration its was essentialy a happy coincidence;-)

PR #589, opened 2020-02-13
Update parameters.py

rieder:
warning->exception

### DIFF
--- a/src/amuse/datamodel/parameters.py
+++ b/src/amuse/datamodel/parameters.py
@@ -307,7 +307,7 @@ class ParametersWithUnitsConverted(object):
 
     def __setattr__(self, name, value):
         if not name in self._original._mapping_from_name_to_definition:
-            raise exceptions.CoreException("Could not set unknown parameter '{0}' for a '{1}' object".format(self.name, type(object).__name__))
+            raise exceptions.CoreException("Could not set unknown parameter '{0}' for a '{1}' object".format(name, type(self._instance()).__name__))
 
         try:
             setattr(self._original, name, self._converter.from_source_to_target(value))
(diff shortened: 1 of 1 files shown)


src/amuse/datamodel/parameters.py
```diff
@@ -308,8 +307,8 @@ def __getattr__(self, name):
 
     def __setattr__(self, name, value):
         if not name in self._original._mapping_from_name_to_definition:
-            warnings.warn("tried to set unknown parameter '{0}' for a '{1}' object".format(name, type(self._instance()).__name__), exceptions.AmuseWarning)
-            return
+            raise exceptions.CoreException("Could not set unknown parameter '{0}' for a '{1}' object".format(self.name, type(object).__name__))
```
ipelupessy: self.name?

ipelupessy: object->self.instance() ?

rieder: I clearly did a bad job copy/pasting and keeping the correct bits... fixing it now.

ipelupessy: ok, above should also be self._original

rieder: where?

ipelupessy: type(object).__name__ -. type(self._original).__name__
I am not sure if thats really informative (it will be some parameter set type name) but at least it 
will work

rieder: ok this is outdated, I changed it back to _instance (as it was in the warning). Should it be _original instead?

ipelupessy: I think so, I am not sure _instance exists, and it is consistent with if statement...

rieder: yes, I think you're right.

Issue #579, opened 2020-02-03
Brutus worker dies


GFTwrt:
Hello,

I get the following message:
amuse.support.exceptions.CodeException: Exception when calling function 'evolve_model', of code 'BrutusInterface', exception was 'Error in code: no error message - code probably died, sorry.'

Condition:
- Debian 10
- git commit 13ca566e32ba2fdd78bae0b33f188b6cf250d52b
   Author: Inti Pelupessy <ipelupessy@users.noreply.github.com>
   Date:   Mon Jan 13 18:25:44 2020 +0100

- small model with 10 bodies
- usage of gravity.set_bs_tolerance_string("1e-20") or even gravity.set_bs_tolerance(1e-20)
With Bs-tolerance 1e-19 I can vary word-lenght from 112 to 130 and even eta in a wide range without crash. Once I use bs_tolerance 1e-20 the worker dies

Thank you fo r the support.

rieder:
@tjardaboekholt could you have a look at this?

tjardaboekholt:
Hi, thanks for the error message concerning Brutus.

I managed to do some runs with e=1e-20, and say, Lw=128 and dt_param=0.10, and the code did not crash. If I reduce Lw to 40 bits, then it crashes and gives the same error message you quoted. This is because the number of bits was too low to reach convergence of e=1e-20. If Brutus fails to reach a converged solution within the maximum number of iterations, it will give up and this causes the code to stop. However, for suitable combinations of (e, Lw, dt_param), the code should in principle work fine, i.e. make sure you have enough bits to resolve e. 

Cheers!

GFTwrt:
Hello,

I identified the values you've written as:
 gravity.set_bs_tolerance_string("1e-20")    # as your "e"
 gravity.set_word_length(130)                    #as your Lw
 gravity.set_eta(0.01)                                  #as your dt_param
Are this the wrong parameters?

GFTwrt:
Hello,

here the build.log. Maybe there is missing something see warnings
Building code: brutus, target: all, in directory: src/amuse/community/brutus
****************************************************************************************************
make[1]: Verzeichnis „/home/pi/amuse/src/amuse/community/brutus“ wird betreten
mpicxx -g -O2 -fPIC  -std=c++0x -I../mpfrc++  -I/home/pi/amuse/lib/stopcond -Impfrc++ -I./src -c -o interface.o interface.cc 
In file included from ./src/Star.h:6,
[...]

Issue #155, opened 2017-07-13
Support for arbitrary precision attributes for (e.g.) Brutus


ipelupessy:
Brutus particle sets currently do not support arbitrary precision..most elegant would be to allow
particle attributes to be mpmath or gmpy arbitrary precision numbers and send those over as strings (using the lowlevel functions that are already implemented for brutus)

this is important to allow saving and resuming sims at arbitrary precision

rieder:
Hi,

This seems related to a problem I had at some point, where I wanted to store lower-precision numbers.
Nathan wrote the following then:

> Quantities in amuse get their data type from their unit. This may sound strange, but it's actually quite natural if you consider enumeration units, string units and particle keys. The default dtype of a unit is indeed a double precision float. This is very easy to override, if you know how to do it ;)

```python
from amuse.units.core import unit_with_specific_dtype
from amuse.lab import *
import numpy
set = Particles(4)
set.mass = [0.01] | unit_with_specific_dtype(units.kg, numpy.float32)
set.radius = [0.9] | unit_with_specific_dtype(units.m, numpy.int32)
print set
                 key         mass       radius
                   -  kg_<type 'numpy.float32'>  m_<type 'numpy.int32'>
====================  ===========  ===========
 3563617830469061953    1.000e-02            0
 1881275734648031197    1.000e-02            0
   93170770887821082    1.000e-02            0
 9144493898677132023    1.000e-02            0
====================  ===========  ===========
```

Maybe this would work similarly for arbitrary precision number particle sets?

ipelupessy:
I am not sure I like the unit_with_specific_dtype (not because currently it is buggy), I think it should be enough that number has a certain dtype...(but maybe it isn't??)..as a side note: enumeration units, string units and particle key units should maybe be deprecated (they are or should no longer be necessary)...

ipelupessy:
its not pretty but:
```
m_= unit_with_specific_dtype(units.m, object)
```
in combination with:
```
a=numpy.array(gmpy.mpf(i) for i in range(10)], dtype=object) | m_
p=Particles(10)
p.x=a
```
substantially works, there are some ugly bits though (unit string= m_object), there may not be more than 1 m_object etc, e.g. copies between inmem sets work, copies to code sets will not work I think...
(not checked)

ipelupessy:
note the comment on #579

PR #597, opened 2020-02-28
mpreal updated; Brutus interface enabled for mpmath.py

GFTwrt:
Hello,

please check and merge brutus interface.cc and interface.py to master to fulfill #155 as mentioned in #579

Actual the mpmath functions do not support units. So I decided to use SI in this functions only. I can add units to mpf-type variables but get errors in next mathematical operations. 

Thank You!
GFTwrt


src/amuse/community/brutus/Makefile
```diff
@@ -9,9 +9,9 @@ AMUSE_DIR?=../../../..
 # mpfr: http://www.mpfr.org/
 # mpfrc++: http://www.holoborodko.com/pavel/mpfr/
 # Change these paths in this Makefile and in src/makefile
-GMP_LIBS   ?= -L/data1/boekholt/Packages/mpreal_local/lib/ -lgmp  # libgmp.a
-MPFR_LIBS  ?= -L/data1/boekholt/Packages/mpreal_local/lib/ -lmpfr # libmpfr.a
-MPFR_FLAGS ?= -I/data1/boekholt/Packages/mpreal_local/include/    # mpreal.h
+GMP_LIBS   = -L/usr/lib/x86_64-linux-gnu/ -lgmp  # libgmp.a
+MPFR_LIBS  = -L/usr/lib/x86_64-linux-gnu/ -lmpfr # libmpfr.a
+MPFR_FLAGS = -I./mpfrc++    # mpreal.h
```
ipelupessy: these should be picked up by the amuse config, so overriding those can break the compilation - why was this necessary?

src/amuse/community/brutus/interface.py
```diff
@@ -182,12 +318,131 @@ def __init__(self, convert_nbody = None, **options):
...
     def initialize_code(self):
         result = self.overridden().initialize_code()
         self.parameters.brutus_output_directory = self.output_directory
         return result
     
+    def get_potential_energy_p_si(self):
+        self.adjust_prec()
+        a=self.convert_nbody.to_si(nbody_system.energy).number
+        b=mpmath.mpf(self.get_potential_energy_string())*a
+        return b #* self.convert_nbody.to_si(nbody_system.energy).unit
```
ipelupessy: I would put this, and the other get_.._p_si in brutusInterface instead; then they should not need the nbody conversion (can be named get_.._mpreal instead)

ipelupessy: also: why is the adjust_precision necessary?

ipelupessy:
@tjardaboekholt can you also comment?

src/amuse/community/brutus/Makefile
GFTwrt: My request was to use use the interface.cc and Interface.py. The makefile is not of interest. changes in the makefile was to get it work on my machine. But need not be relevant for the master. I did not know that I can't choose the files for pull request sepaerartely.

ipelupessy: ok, no problem will fix it..
I don't think you can select files from a pull request, usually I make sure I only commit relevant files in a seperate branch, eg:
```
git checkout master
git branch brutus_mpreal
git checkout brutus_mpreal
git add <file1> <file2> etc
git commit
git push <your remote> brutus_mpreal:brutus_mpreal
```
(a seperate branch is handy since it will allow you to work on multiple things at the same time)

src/amuse/community/brutus/interface.py
GFTwrt: Hello,
you have recognized "import mpmath"? The actual Brutusinterface is not able to handle mpmath. If you want to integrate a complete mpmath handling to your interface I'd welcome it. But for the moment I can't do this. Therefore I put the conversion to the driver to not impact all parts of the system. The adjustment is nessesary to get same precision on both sides of the interface ( Python-mpmath; C++-mpreal)

ipelupessy: yea, I saw it...I think its ok for now..I am not a 100% sure how mpmath manages the precision in its context..it could also be put in the setter of the precision of Brutus and/or commit_parameters..but don't worry about that..

GFTwrt:
The goal of this is to get the same precision on code and python side. The actual numbers are only 64-bit-floats. But this is not enough for my application. I need at least 50 significant figures. Brutus itself can handle this but the interface was not able to do so. So the only way is to convert mpreal to string and send it as characters trough the interface. At the python side mpmath collects the characters again and converts it to as number again. Mpmath as well as mpmath needs to be configured according to the amount of significant figures.

src/amuse/community/brutus/interface.py
GFTwrt: mpmath allows you to increase and decrease see  http://mpmath.org/doc/current/basics.html?highlight=prec#temporarily-changing-the-precision. But I do just increase. The other direction is not of interest. I think this more easy than handling different setters.

tjardaboekholt:
I did the same as  GFTwrt's last comment, i.e. request the full string from Brutus, and then convert it to mpmath type on the Python side.This in principle should work okay. Alternatively, we can let the interface do this for us.

I found I needed mpmath only if I am interested in the growth of very small perturbations. In other cases where I am interested in the converged solution to the first few (say 6) decimal places, then I do not require all the characters Brutus has, and requesting a double suffices.

GFTwrt:
Hello,
please keep in mind: I haven't deactivated anything. -> The standard interface works like before. But you're able to get the whole precision additionally.

ipelupessy:
ok, I have added some code to make sure everything works if mpmath is not installed - making mpmath a optional dependency..furthermore:
- the PR needs some tests to test the new function (to be added to test/codes_tests/test_brutus.py, it passes current tests so thats ok;-) 0
- id is a python keyword, wherever its used as index argument, it should be replaced by "index"
- i think the _p_si should be moved to BrutusInterface and the unit conversion not in there (thus the name is not ok, should be then .._mpreal (keep in mind you would want to be able to use mpreals also when convert_nbody=None)
- if you want to keep the convenience functions rewrite them to use the lowlevel .._mpreal

GFTwrt:
Hello, didn't you want to merge?

Please check my latest change in the main Makefile to pull too.

ipelupessy:
is it possible to address the 3 issues above (tests and the two regarding the _p_si ? at leasts the tests are very important (the others I can have a look at later, but then its good to have the tests to check nothing breaks...)

GFTwrt:
Can't follow you. What tests are you expecting? This is additional stuff. If the std-automated-Test for brutus works it should be fine. The rest of the interface just delivers a higher count of figures. To bring the things to the normal interface would mean to change the whole interface with all python files to enable "mpmath.py". The naming is p_si and not mpreal, because mpreal is just one implementation of multiprecision (see: https://www.boost.org/doc/libs/1_72_0/libs/multiprecision/doc/html/boost_multiprecision/intro.html  or http://homepages.laas.fr/mmjoldes/campary/ or http://mpmath.org/doc/current/index.html or https://escholarship.org/uc/item/69q5t2mj ...). The "p" is the synonym for "precise". "si" stands for "Système international d’unités". My actual goal is the usage of brutus as a reference for a mp-gadget2 using CAMPARY in BOOST MP-Template evironment to enable my solarwind simulations in our solar system. The precision is needed, because my calculations show that the power of the solar wind is in dimensions of milli Watt, while the total energy is in ranges of 10³⁵ Joule.

ipelupessy:
the new interface functions should be included in the test suite; you can look at test/codes_tests/test_brutus.py, then add a few tests which check that the _p_si methods work (by e.g. setting and retrieving random values)..its to ensure they keep on working in the future if someone works on the brutus interface..

GFTwrt:
test/codes_tests/test_brutus.py just defines classes. Where are they called?

ipelupessy:
normally you call them through nosetest:
```
mpiexec nosetests test/codes_tests/test_brutus.py -sv
```
(-sv gives verbose output) will try to run all tests in the file, you can specify 1 test,  e.g.:
```
mpiexec nosetests test/codes_tests/test_brutus.py:TestBrutusInterface.test1 -sv
```

GFTwrt:
Hello, I'm a little confused about the actual test. I'd interpret it as fail. STDerr looks fine but STDerr is confusing me. Could you please have a look to both channel and comment after.
[stderr.log](https://github.com/amusecode/amuse/files/4372076/stderr.log)
[stdout.log](https://github.com/amusecode/amuse/files/4372079/stdout.log)

GFTwrt:
Just nosetests seems to work.
[just_nosetest.log](https://github.com/amusecode/amuse/files/4372157/just_nosetest.log)

GFTwrt:
Please check test.

ipelupessy:
ok, thanks!

Issue #603, opened 2020-03-18
NameError: name 'limepy' is not defined in amuse.ic.limepy.new_limepy_model 


tlrh314:
**Describe the bug**
The method `amuse.ic.limepy.new_limepy_model` raises NameError caused by `ImportError: cannot import name 'PiecewisePolynomial' from 'scipy.interpolate'` in `amuse.ic._limepy/limepy.py`

```txt
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-25-c742fd27ed6b> in <module>
     14 
     15 
---> 16 p = new_limepy_model(7, 1)
     17 
     18 converter = nbody_system.nbody_to_si(1e5 | units.MSun, 40 | units.parsec)

bla/lib/python3.8/site-packages/amuse/ic/limepy.py in new_limepy_model(*args, **kwargs)
     55     conv = kwargs.pop("converter", None)
     56 
---> 57     l = Limepy(*args, **kwargs)
     58     p = l.result
     59 

bla/lib/python3.8/site-packages/amuse/ic/limepy.py in __init__(self, *args, **kwargs)
     32         kwargs["scale_radius"] = "rv"
     33 
---> 34         self.model = limepy(*args, **kwargs)
     35         self.kwargs = kwargs
     36 

NameError: name 'limepy' is not defined
```

```txt
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-2-5e92325257de> in <module>
      6 from amuse.units import units
      7 from amuse.units import nbody_system
----> 8 from amuse.ic.limepy import new_limepy_model
      9 
     10 

bla/lib/python3.8/site-packages/amuse/ic/limepy.py in <module>
      7 
      8 try:
----> 9     from ._limepy import limepy, sample
     10     scipy_imported = True
     11 except ImportError:

bla/site-packages/amuse/ic/_limepy/__init__.py in <module>
      5 __version__ = '0.1.0'
      6 
----> 7 from .limepy import limepy
      8 from .sample import sample
      9 

bla/lib/python3.8/site-packages/amuse/ic/_limepy/limepy.py in <module>
      4 import scipy
      5 from numpy import exp, sqrt, pi, sin
----> 6 from scipy.interpolate import PiecewisePolynomial, interp1d
      7 from scipy.special import gamma, gammainc, dawsn, hyp1f1
      8 from scipy.integrate import ode, quad, simps

[...]

PR #604, opened 2020-03-18
Fix for issue #603: bump limepy to 0.1.1

tlrh314:
Fix for issue #603


rieder (approved):

rieder:
For future reference: we really should just download the correct version of Limepy here instead of duplicating the source code.

PR #620, opened 2020-04-23
New file with functions to convert to/from Astropy

### LINKED ISSUE #170: Astropy units compatibility / conversion?
Since Astropy has a unit systems different from the one in AMUSE while it's quite plausible both packages are used in the same project, it seems practical to make sure these unit systems are in some way compatible. 
The implementations are somewhat different (e.g. Astropy uses the multiplication operator '*' to assign units, where Amuse uses '|', among others), but not completely.

I think it would be useful to check:
- if we use the same definition of units (we really should)
- if we can make an easy conversion module ('astropy_to_amuse' and VV) to replace one unit type with the other for better compatibility
- if it would make any sense at all to switch unit systems (this would probably be a lot of work, so I doubt it's worthwhile unless we do a big overhaul)

A conversion module should be relatively straightforward, using a table of Amuse units and their Astropy equivalent.

rieder:
Currently the conversions are implemented as functions, but perhaps we should also make to_astropy available as a method for AMUSE units?

This fixes #170

### DIFF
--- a/src/amuse/units/astropy.py
+++ b/src/amuse/units/astropy.py
@@ -1,6 +1,5 @@
 """
-This file contains conversion functions to/from Astropy, as well as constants
-taken from Astropy.
+This file contains conversion functions to/from Astropy.
 
 Requires astropy to be installed.
 """
@@ -78,6 +77,3 @@ def to_astropy(quantity):
             ap_quantity = ap_quantity * apu.cd**base_unit[0]
 
     return ap_quantity
-
-
-G = from_astropy(apc.G)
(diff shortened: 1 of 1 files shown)


ipelupessy:
maybe add a basic test?

ipelupessy:
yea, you can skip if astropy is not available...maybe also  add some graceful failure to the astropy import...

rieder:
Yeah I'll do that.
I also added a similar PR to Astropy: https://github.com/astropy/astropy/pull/10196

rieder:
What should a test for this contain?

ipelupessy:
just basic functionality stuff...maybe convert some random quantities to astropy and back and check result...

rieder:
Closing in favour of #687 which probably requires a broader solution.

PR #625, opened 2020-04-28
enforce dtype on quantity

ipelupessy:
I had this lying around..it seems sensible, but has a performance penalty

### DIFF
--- a/src/amuse/units/quantities.py
+++ b/src/amuse/units/quantities.py
@@ -10,7 +10,7 @@ from amuse.support.core import late
 from amuse.support.core import compare_version_strings
 from amuse.units import core
 from amuse.units.si import none
-from amuse.units.core import zero_unit:
+from amuse.units.core import zero_unit
 
 
 """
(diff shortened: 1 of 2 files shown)


rieder:
Definitely seems sensible

rieder:
Codefactor was complaining about some issues that weren't changed in this PR. Not sure why, but I fixed most of these. 
I'm happy for this to be merged. I doubt the performance penalty would be large and I think it's better to enforce the dtypes.

src/amuse/units/quantities.py
```diff
@@ -10,8 +10,7 @@
 from amuse.support.core import compare_version_strings
 from amuse.units import core
 from amuse.units.si import none
-from amuse.units.core import zero_unit
-
+from amuse.units.core import zero_unit:
```
ipelupessy, line 13: colon?

ipelupessy:
the dtype thing eventually has to be reworked ..e.g. right now its not conisstent what the result of the operations with dtypes is...also I think maybe dtype should move to quantity...(in any case, all this is beyond the scope of this PR)

src/amuse/units/quantities.py
rieder: typo I think. will remove.

PR #626, opened 2020-05-01
Docu update

MJCWilhelm:
I've worked on updating the AMUSE documentation, testing, updating to python3, and expanding the gravitational dynamics interface example to discuss some subjects missing from the c++/Fortran tutorials (the biggest addition). Also, I tried to do the Blender tutorial, but couldn't get it to work; do we still want to keep it? I removed it from the index for now.

Additionally, I was still uncertain about some stuff in the definition of parameters. What exactly does the must_set_before_get keyword do? And when are the default_values set, exactly?

Finally, are there aspects to the interface that are still missing? We don't currently have an example for a grid, for one...


rieder:
I'd like to merge this, but I'd prefer to read it as a "live" version first (reading rst files is annoying). I'll see if I can get that working.

PR #634, opened 2020-06-02
separate standalone amuse-units package

rieder:


rieder:
installing amuse-units and then amuse-framework seems to work for me

ipelupessy:
ok, but I think amuse units component themselves need some work to remove e.g. the amuse exceptions etc

ipelupessy:
I think the exceptions can be converted to normal exceptions, however there is also some support utilities used (late, memoization

ipelupessy:
i think late does what functools cached_property (introduced in 3.8) does and the memoization is lru_cache (3.3 or so)

rieder:
Ok. I'm not yet happy with a 3.8 dependency but 3.3 is certainly fine.

ipelupessy:
memoization speeds up the unit module considerably, the late stuff I have never really seen what it adds though

rieder:
> memoization speeds up the unit module considerably

Ah, I see. I guess it should have been memorize :). I'll see if I can replace that.

rieder:
lru_cache is not a simple drop-in replacement unfortunately... needs a bit more work.

rieder:
here's a description of how lru_cache could work for methods... not sure I like this. https://stackoverflow.com/questions/33672412/python-functools-lru-cache-with-class-methods-release-object

ipelupessy:
just copying the necessary bits is always an option...

ipelupessy:
(ie just putting the late and memoization bits also in units) with a comment and view of replacing them with standard constructs...

ipelupessy:
keep open

Issue #425, opened 2019-05-21
OSX travis test errors


ipelupessy:
some tests give the following errors:

The Open MPI wrapper compiler was unable to find the specified compiler
x86_64-apple-darwin13.4.0-clang in your PATH.
Note that this compiler was either specified at configure time or in
one of several possible environment variables.

travis still reports job ok though (which is also wrong)

ipelupessy:
travis retired

Issue #414, opened 2019-05-07
propose move tests to amuse/test/suite/


ipelupessy:
in the binary install this is where they end up anyway

it also fixes an annoying issue with tests with generated python executables
(they need careful set up of the environment, since test/compile_tests is normally not part of PYTHONPATH)

rieder:
sure, why not.

rieder:
Fixed by #637

Issue #351, opened 2018-12-08
Travis openmpi/linux fails


rieder:
**Describe the bug**
OpenMPI/Linux tests fail on Travis, not exactly sure why.

#350 disables these tests for now, since in practice it works fine and it seems better to get at least some useful feedback from testing (instead of them always failing).

ipelupessy:
it used to work, but then something changed on the test machines/configurations (you can go back in history to retry one of the commits that passed and see them fail), so I was hoping it catch it working again;-) in any case maybe better to set allow_failure for this test?

Can you confirm the test works on real machine with the same config??

there is another problem with the travis script, it passes sometimes while tests are failing...

rieder:
Need to test this with the updated config

ipelupessy:
travis retired

PR #637, opened 2020-06-04
Updates to testing

### LINKED ISSUE #631: Travis testing isn't working for Linux/OpenMPI and macOS
Travis testing should work for all cases. I made a new branch to work on this.

rieder:
- overhaul of travis config (for now only linux/mpich, but much easier to extend)
- pytest instead of nose
- move tests to amuse namespace
- some fixes to errors
- some fixes for deprecation warnings

Fixes #631, #414

Needs testing with #425, #351


rieder:
I guess codecov is complaining since the whole test directory is gone now... that can be ignored.

rieder:
@ipelupessy ok to merge?

src/amuse/io/nemobin.py
```diff
@@ -273,147 +282,141 @@ def get_item_header(self):
...
-        
-    
+        self.file.write(x.tobytes())
+
     def write_array(self, typetag, data):
         x = array.array(typetag, data)
         x.append(0)
-        self.file.write(x.tostring())
-        
+        self.file.write(x.tobytes())
```
ipelupessy, line 334: why is this?

rieder: fixes a deprecation warning

src/amuse/support/methods.py
```diff
@@ -1,24 +1,27 @@
 from amuse.support.core import late
 from amuse.support import exceptions
-from amuse.units import nbody_system
+# from amuse.units import nbody_system
```
ipelupessy, line 3: if its not needed, remove?

src/amuse/io/nemobin.py
rieder: (or at least some sort of warning that came up... very specific in that it should be fixed this way.)

src/amuse/support/methods.py
rieder: ya

src/amuse/support/options.py
```diff
@@ -90,137 +90,135 @@ def homedirectory(self):
...
     def override_value_for_option(self, name, value):
         self.overriden_options[name] = value
-        
 
     def read_from_ini_string(self, string):
         file = StringIO(string)
-        self.config.readfp(file)
-    
-    
+        self.config.read_file(file)
```
ipelupessy, line 126: confirm this is ok?

rieder: yes - again a deprecation warning with specific fix listed

src/amuse/test/suite/compile_tests/test_c_implementation_simplified.py
```diff
@@ -597,6 +598,7 @@ def test30(self):
           "got multiple values for argument 'i' of method dummy_3_int")
         instance.stop()
 
+    @pytest.mark.skip
```
ipelupessy, line 601: hmmm. this test should work...

src/amuse/test/suite/compile_tests/test_c_implementation_simplified.py
```diff
@@ -1,3 +1,4 @@
+import pytest
```
ipelupessy, line 1: I think its better to use the unittest skip

src/amuse/test/suite/compile_tests/test_c_implementation_simplified.py
rieder: Yes - but it doesn't... same for test31 in fortran implementation.

src/amuse/test/suite/compile_tests/test_fortran_implementation.py
```diff
@@ -1,3 +1,4 @@
+import pytest
```
ipelupessy, line 1: same as above

src/amuse/test/suite/compile_tests/test_c_implementation_simplified.py
rieder: it's best not to skip but sure :)

src/amuse/test/suite/compile_tests/test_stopping_conditions.py
```diff
@@ -573,7 +573,7 @@ def get_mpidir(cls):
         
     @classmethod
     def get_codestring(cls):
-        return CodeStringF
+        return codestringF
```
ipelupessy, line 576: so is pytest still testing this class (it shouldn't)?

src/amuse/test/suite/compile_tests/test_c_implementation_simplified.py
rieder: no idea why yet, but these tests hang.

src/amuse/test/suite/compile_tests/test_stopping_conditions.py
rieder: yes, this came up as an error.

src/amuse/test/suite/core_tests/test_attribute_storage.py
```diff
@@ -10,55 +10,55 @@
...
 
 class _AbstractTestInMemoryAttributeStorage(amusetest.TestCase):
-    
+
     def new_inmemory_storage(self, is_with_units = True):
         raise NotImplementedError()
-        
-    def test1(self):
+
+    def xtest1(self):
```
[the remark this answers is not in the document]
rieder: my bad - this was me disabling them temporarily. will revert.

src/amuse/test/suite/core_tests/test_store.py
```diff
@@ -28,7 +29,8 @@ def store_version(self):
         
     def get_version_in_store(self, container):
         raise NotImplementedError
-        
+
+    @pytest.mark.skip
```
ipelupessy, line 33: don't think this should be the solution...

src/amuse/test/suite/core_tests/test_attribute_storage.py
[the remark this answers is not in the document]
rieder: https://docs.pytest.org/en/latest/goodpractices.html#conventions-for-python-test-discovery

src/amuse/test/suite/core_tests/test_store.py
rieder: no

src/amuse/units/generic_unit_converter.py
```diff
@@ -113,15 +113,15 @@ def matrixrank(self, A, tol=1e-8):
         return numpy.sum(numpy.where( s>tol, 1, 0 ) )
 
     def determine_new_base(self):
-        matrix = numpy.mat(numpy.zeros((self.system_rank,self.system_rank)))
+        matrix = numpy.asmatrix(numpy.zeros((self.system_rank,self.system_rank)))
```
ipelupessy, line 116: I think pytest is warning that matrix is deprecated? should be ordinary array and then the matrix operations should be replaced ( * -> .dot( and **-1 -> numpy.linalg.inv())

rieder: ya. sounds good.

rieder: matrix is not yet deprecated but "pending deprecation" and not recommended. 
asmatrix and mat are the same, I found asmatrix easier to find when it needs to be replaced.

src/amuse/test/suite/core_tests/test_attribute_storage.py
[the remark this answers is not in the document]
ipelupessy: hmmm...https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure

src/amuse/test/suite/compile_tests/test_stopping_conditions.py
ipelupessy: the change is ok, but it shouldn't test the class

src/amuse/test/suite/core_tests/test_store.py
rieder: how about adding an is_abstract method to the class, skip if it returns True?

rieder: seems clumsy but I can't figure out a better method...

rieder: never mind, can just add @unittest.skip() to a class

rieder: but then that gets inherited... brrrr....

ipelupessy:
I am still testing and running into issues (which may be my config)..

rieder:
ok - travis runs ok apart from the issue with test_attribute_storage's setup class not being skipped...

rieder:
(and the test31 issues)

ipelupessy:
the last commit fixes the case where baseclass is tested (when it shouldn't). Apperently pytest also tries all classes derived from UnitTestCase.. (not sure if that is documented). Can be solved by making it mixin

ipelupessy:
so you can also fix the other cases this way...

src/amuse/units/generic_unit_converter.py
rieder: this is not immediately necessary so I'll leave it for now

ipelupessy:
have you tested the codes test?

ipelupessy:
there are some codes test which leave debris in the cwd - capreole and and maybe others..setup and teardown with chdir can be added for those...

rieder:
@ipelupessy That timeout you removed was a timeout override, allowing more time for that specific tests... I guess we could allow more time in general but I'd rather not do that and keep the override instead.

src/amuse/test/suite/compile_tests/test_c_implementation_simplified.py
rieder: it should, but it doesn't. test31 in test_c_implementation_simplified.py and test_fortran_implementation is problematic for some reason, both on travis and for me locally (macOS + Linux). I don't know why...

ipelupessy:
ok, I had some issues with a pytest not supporting this...

rieder:
yes, it's a plugin for pytest (pip install pytest-timeout)

rieder:
but I'm just allowing more time for core-tests now, should be ok (these don't take long generally)

rieder:
I'll merge if you're ok with it. Some tests need more addressing, but that's the same for the old testing setup...

rieder:
@ipelupessy ok?

ipelupessy:
ok

PR #645, opened 2020-06-15
Also fix hasing for gas_interface.py in GalactICs

rknop:
Same change Inti made to in 2b632905775c67f04f369b8183a12b0b8dae30f9 for GalactICs interface.py to fix hashing in Python 3, now also in gas_interface.py.


ipelupessy:
oops, forgot about this, thanks!

PR #647, opened 2020-06-16
add openmpi oob env var

ipelupessy:


ipelupessy:
this is to test the ci with github actions w/o commiting every single change to master

PR #655, opened 2020-06-29
Updates to brokenimf

### LINKED ISSUE #653: mass_min support for other IMFs than Salpeter
The new_salpeter_mass_distribution function in AMUSE has support for a mass_min keyword, while other IMFs (e.g. those in brokenimf.py) don't support this. 
We may want to make this keyword more universal.

rieder:
This will fix #653

### DIFF
--- a/src/amuse/ic/brokenimf.py
+++ b/src/amuse/ic/brokenimf.py
@@ -11,16 +11,28 @@ import numpy.random
 from amuse.units import units
 from amuse.units.quantities import AdaptingVectorQuantity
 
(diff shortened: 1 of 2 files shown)


rieder:
(work in progress)

rieder:
I think this works now - tested it for myself...

rieder:
This also fixes a possible issue for very low values of mass_max, i.e. where mass_max would be smaller than not just the last bin.

rieder:
added

Issue #553, opened 2019-12-17
Generalised tests


rieder:
Currently we have tests for all individual community codes. These replicate large portions of each other, especially codes in the same domain.
It would be good to have generalised tests available that can run for all codes (within a domain).
These could include basic tests (does a code initialise/stop/set parameters), but perhaps also sanity checks (is the result like expected). 
If this is possible, the latter of these should probably not result in "hard" fails, but in warnings (e.g. "do not use this code for this kind of problem"). Alternatively, the test could be overridden by a specific code test, or ignored for a specific code.

PR #657, opened 2020-07-02
Patch generalised code testing

rieder:
Attempt at fixing #553


ipelupessy:
I think in principle you could do the mixin trick we did previously, so not have the base test class derive from TestWithMPI but only the concrete test classes (testHermite etc)?

rieder:
Thanks, that works!

PR #660, opened 2020-07-16
Update gd.py to reduce recommit_particles calls

rieder:
Allow set_* methods from UPDATE state without going through RUN state, reducing the amount of recommit_particles calls

### DIFF
--- a/src/amuse/support/state.py
+++ b/src/amuse/support/state.py
@@ -112,7 +112,7 @@ class StateTransition(object):
         self.handler = handler
 
     def __str__(self):
-        return "transition from {0} to {1}".format(self.from_state, self.to_state)
+        return "{0}: transition from {1} to {2}".format(self.handler.interface.model_name,self.from_state, self.to_state)
 
     def do(self):
         if self.handler.log_transitions:
@@ -142,7 +142,7 @@ class StateMachine(OptionalAttributes):
         
     @option(type='boolean', sections=['state',])
     def log_transitions(self):
-        return False
+        return True
         
     def enable(self):
         self.is_enabled = True
--- a/src/amuse/support/state.py
+++ b/src/amuse/support/state.py
@@ -115,8 +115,7 @@ class StateTransition(object):
         return "{0}: transition from {1} to {2}".format(self.handler.interface.model_name,self.from_state, self.to_state)
 
     def do(self):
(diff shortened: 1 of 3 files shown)


rieder:
Test below should have only one call to recommit_particles, before this patch there are two.

```python
import logging
from amuse.lab import *

logging.basicConfig(level=logging.DEBUG)
logging.getLogger("code").setLevel(logging.DEBUG)

particles = new_plummer_sphere(100)

grav = Hermite(redirection="none")
grav.particles.add_particles(particles)
grav.evolve_model(0.01 | nbody_system.time)

channel = particles.new_channel_to(grav.particles)
grav_attributes = ["x", "vx"]

particles.x += 1 | nbody.length
particles.vx += 1 | nbody.speed

channel.copy_attributes(grav_attributes)
grav.evolve_model(0.02 | nbody_system.time)

grav.stop()
```

ipelupessy:
hmm, the script you provide is not working here (doesn't want to set "x" and "vx" only)

ipelupessy:
I have enabled/added logging; you can add more logging in the state if you wish..

rieder:
> hmm, the script you provide is not working here (doesn't want to set "x" and "vx" only)

Odd. I initially tested with xyz and vxyz, changed it now to do that again (I thought just x and y would simplify it, weird that it doesn't work!).

ipelupessy:
its a limitation from the fact that set_position sets all three position attributes at once...its a bit weird that "position" also doesn't work...

ipelupessy:
I think its ok to merge btw, there may be other problems (it could use a bit more tests), but with the better logging hopefully it will be easier to resolve these...

rieder:
position is a read_only version of [x,y,z] I think... so it can't be set.

src/amuse/support/state.py
```diff
@@ -139,11 +138,7 @@ def __init__(self, interface, **options):
     @option(type='boolean', sections=['state',])
     def is_enabled(self):
         return True
-        
-    @option(type='boolean', sections=['state',])
```
rieder, line 143: why remove this? maybe just default to True instead?

ipelupessy: its just duplicating logging functionality..we also don't do it when logging the channel stuff...

Issue #670, opened 2020-09-02
Let Fi kick without needing a timestep


rieder:
> yea, it initializes always, necessitating a timestep shift (maybe this could be turned off)

_Originally posted by @ipelupessy in https://github.com/amusecode/amuse/issues/669#issuecomment-683074233_

Fi currently can be used as a "kick" code in Bridge, but it will always calculate a timestep. Occasionally this timestep can be considered "too small", crashing the code. 

For use as a "kick" code, this timestep isn't needed so it would be nice if it wasn't calculated at initialize time, only when needed (i.e. at 'evolve' time).

rieder:
I'm guessing this is done in the `inittimestep` call in `muse_finalize_init`, which itself is called in `commit_particles`. Maybe `inittimestep` can be moved to `evolve_model`?
Or is it needed earlier?

ipelupessy:
no, its the postprocessread which triggers an refresh_itimestep, which has the offending timestep check (and somewhere along the way it does the timeconsuming force calc)....

ipelupessy:
..however this bit in interface.py:
```
        handler.add_method('EDIT', 'get_gravity_at_point')
        handler.add_method('EDIT', 'get_potential_at_point')
```
indicates that get_gravity and get_potential can be called in EDIT (and thus not need commit_particles) - I think it should be possible to do bridge w/o triggering expensive initialization already... make you can try to track down why it does the commit_particles?

ipelupessy:
see also commit message https://github.com/amusecode/amuse/commit/9dbca60f77ad047955dea4bdfc3da367355d960b

rieder:
hmm, a quick test suggests that this is indeed possible and that get_gravity_at_point should work in EDIT state. I wonder why this doesn't happen in Bridge in my simulation currently. Investigating further...

rieder:
stranger still, Fi seems to remain in EDIT. So why is this 'timestep' thing happening then...

rieder:
So this is the typical output from Fi at the moment it crashes/hangs:
```
Creating field tree code
  --- FI Amuse interface ---
  omp threads:          16
recommit_particles
  timestep too small
STOP aborting
```
while debug logging tells me that Fi is still in EDIT:
```
20200902 17:48:44 - state - INFO: Fi: transition from state 'INITIALIZED' to state 'EDIT'
```

rieder:
it's being called from `muse_finalize_init`.

rieder:
[...]

PR #671, opened 2020-09-02
remove calls to commit_particles in bridge.py

rieder:
See #670 - commit_particles calls shouldn't happen here, the state model will take care of this.


rieder:
If merged, this would close #670

ipelupessy:
yes, I agree!

ipelupessy (approved):

ipelupessy:
the original commit doesn't mention any reason for adding the commit_particles, it was probably because the state model was a bit flaky at the time..

Issue #643, opened 2020-06-12
Decouple amuse and seba repositories


rieder:
The main repository for SeBa development is currently the same as the main AMUSE repository. 
Since both are actively developed and largely independent of each other, I think it would be better to decouple these repositories (and use a download script to get SeBa for AMUSE). That way both can have distinct versions with clear change logs. Also would make it clearer who is responsible for each part.
This should be possible while retaining the file histories.

rieder:
@silviatoonen @ipelupessy I hope to hear your thoughts on this.

ipelupessy:
I have no preference either way..

rieder:
it can be done using 'git subtree split'. I made a splitoff in https://github.com/rieder/seba to see how this would work.

ipelupessy:
btw, seba build is currently broken

ipelupessy:
and there are a lot of warnings of the "no return statement in function returning non-void" type. Despite being just a warning this can cause nasty seg fault errors.

ipelupessy:
for example:
```
#include <stdio.h>

int f()
{
}

int main()
{
  printf("calling returnless nonvoid\n");
  f();
  return 0;
}
```
compiling this with:
```g++ test.c```
and then running segfaults on recent gcc

spzwart:
Steven,
I am not in favor of decoupling SeBa from AMUSE-framework.
The reason is that I think that AMUSE should come with a full suit of minimal codes, and I would count SeBa as the fundamental stellar/binary evolution code in AMUSE.

Simon

rieder:
I fully agree on that point, and I don't think this is at risk by decoupling the repositories.
Not even with the "developer" installation - we can use git subtree (https://www.atlassian.com/git/tutorials/git-subtree) to link repositories.

ipelupessy:
it does complicate a little bit...the earlier git submodules where not so nice in this respect, don't know about subtrees..

rieder:
subtree is supposed to be a lot better in this respect. Still, requires some testing to find out exactly.

rieder:
> and there are a lot of warnings of the "no return statement in function returning non-void" type. Despite being just a warning this can cause nasty seg fault errors.

@silviatoonen did you notice this?

silviatoonen:
If the subtree option works well, I'd be in favour of that. It's been a bit messy previously with the Amuse development and the SeBa development through one another. 
Ps I've fixed the bug in SeBa. 
[...]

PR #684, opened 2020-09-29
Move SeBa to its own repository

rieder:
See #643


ipelupessy:
the download script still has the --petar-version flag; other than that it looks ok...will the packaging still work though?

rieder:
Yes, a package can still be made this way. See the amuse-petar package, which already works this way.
May need some updating before it works though - I didn't test this yet.

PR #691, opened 2020-10-26
Add $(NVCC_FLAGS) to a bunch of Makefiles

rknop:
Several of the Makefiles on the community codes weren't passing $(NVCC_FLAGS) along to the nvcc compiler.  I've added this to those Makefiles.


rknop:
D'oh... broke higpus build.  Will fix that, create a follow-up pull request.

rieder:
Looks alright to me, but I can't check since I don't have a GPU machine currently. @spzwart could you verify these all build for you?

PR #692, opened 2020-10-27
Add more documentation to amuse.ic package

rieder:
Also make molecular_cloud and solarsystem accessible via amuse.ic

### DIFF
--- a/src/amuse/__init__.py
+++ b/src/amuse/__init__.py
@@ -16,9 +16,9 @@ community - a variety of existing codes of different physical domains, each with
             a uniform interface to enable coupling
 
 Help is available for each of these packages, e.g.:
-> amuse.sh
->>> import amuse.ic
->>> help(amuse.ic) # doctest: +ELLIPSIS
+> ipython
+In [1]: import amuse.ic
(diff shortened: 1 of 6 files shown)


rieder:
@amusecode/core please check this is OK.

src/amuse/__init__.py
```diff
@@ -16,9 +16,9 @@
             a uniform interface to enable coupling
 
 Help is available for each of these packages, e.g.:
-> amuse.sh
->>> import amuse.ic
->>> help(amuse.ic) # doctest: +ELLIPSIS
+> ipython
+In [1]: import amuse.ic
+In [2]: help(amuse.ic) # doctest: +ELLIPSIS
```
ipelupessy, line 21: I don't think its necessary to use ipython for this...

rieder: No, but it's just an example. Python works fine too. Just wanted to get rid of amuse.sh which we shouldn't be using anymore.

rieder: fixed anyway.

ipelupessy:
sure..

Issue #687, opened 2020-10-17
Problem with using pipe operator for units in py3.9+


rieder:
Python 3.9 and 3.10 have adopted PEPs 584 (1) and 604 (2) respectively. These use the pipe operator, and this might very well collide with our use of the pipe operator for creating quantities.

We will need to find a solution for this relatively soon. 

1: https://www.python.org/dev/peps/pep-0584/
2: https://www.python.org/dev/peps/pep-0604/

rieder:
One obvious solution would be to adopt Astropy's system of units. We have discussed this before and it seems like this would be another reason to do so. 
This is of course not ideal for other *muse projects, as we don't want them to depend on astropy.
A hybrid solution could also be an option, where we move to the * operator for creating quantities but still use our own unit system. I'm unsure about the consequences this would have for distinguishing quantities and units though (how is this done in Astropy?).

spzwart:
Sounds like a perfect topic for our workshop in the last week of October.

Simon

PR #693, opened 2020-10-28
Attempt at replacing | with * for creating quantities

rieder:
See issue #687. This PR makes it possible to construct quantities using the `*` operator, which would previously have resulted in a unit.
With this PR, a unit multiplied/divided by another unit will still give a unit, while other multiplications/divisions will result in a quantity. Of course, quantities can still be converted to units with `.to_unit()`.

### DIFF
--- a/src/amuse/units/core.py
+++ b/src/amuse/units/core.py
@@ -7,28 +7,29 @@ from amuse.support.core import memoize
 
 from amuse.support.core import MultitonMetaClass
 
+
 class system(object):
     ALL = {}
-    
+
     def __init__(self, name):
         self.name = name
         self.bases = []
         self.mapping_from_base_name_to_base = {}
         self.ALL[self.name] = self
         self.index = len(self.ALL)
-    
+
     def reference_string(self):
         return "{0}.get({1!r})".format('system', self.name)
-            
+
     def add_base(self, unit):
         unit.system = self
         unit.index = len(self.bases)
         self.bases.append(unit)
         self.mapping_from_base_name_to_base[unit.quantity] = unit
-    
+
     def base(self, name):
         return self.mapping_from_base_name_to_base[name]
-        
+
     @classmethod
     def get(cls, name):
         try:
@@ -37,29 +38,27 @@ class system(object):
             from amuse.units import nbody_system
             from amuse.units import si
             return cls.ALL[name]
-            
 
     def __reduce__(self):
         return (get_system_with_name, (self.name,))
-    
-    
+
     def __str__(self):
         return self.name
-            
+
 
 class unit(object):
     """
     Abstract base class for unit objects.
-        
+
     Two classes of units are defined:
 
     base units
         The base units in a given system of units. For SI, these
-        are meter, kilogram, second, ampere, kelvin, mole and 
+        are meter, kilogram, second, ampere, kelvin, mole and
         candele. See the si module :mod:`amuse.units.si`
     derived units
         Derived units are created by dividing or multiplying
-        with a number or with another unit. For example, 
+        with a number or with another unit. For example,
         to get a velocity unit we can devine vel = 1000 * m / s
 
     Units can also be named, by creating a named unit.
@@ -67,60 +66,65 @@ class unit(object):
     __array_priority__ = 100
 
     def __mul__(self, other):
-       
+
         if isinstance(other, unit):
             return mul_unit(self, other)
         else:
-            return other*self
+            return self.new_quantity(other)
+            # return other*self
 #            return factor_unit(other, self)
-        
+
     def __truediv__(self, other):
         if isinstance(other, unit):
             return div_unit(self, other)
         else:
-            return (1.0/other)*self
+            return self.new_quantity(1.0/other)
+            # return (1.0/other)*self
 #            return factor_unit(1.0 / other, self)
 
     def __rmul__(self, other):
         if other == 1:
             return self
         else:
-            return factor_unit(other, self)
-    
+            if isinstance(other, unit):
+                return factor_unit(other, self)
+            return self.new_quantity(other)
+
     def __ror__(self, value):
         """Create a new Quantity object.
-    
-        :argument value: numeric value of the quantity, can be 
+
+        :argument value: numeric value of the quantity, can be
             a number or a sequence (list or ndarray)
-        :returns: new ScalarQuantity or VectorQuantity object 
+        :returns: new ScalarQuantity or VectorQuantity object
             with this unit
-            
+
         Examples
-        
+
         >>> from amuse.units import units
-        >>> 100 | units.kg
+        >>> 100 * units.kg
         quantity<100 kg>
         """
         return self.new_quantity(value)
-        
+
     def __rtruediv__(self, other):
-        return factor_unit(other, pow_unit(-1,self))
-    
+        # return factor_unit(other, pow_unit(-1, self))
+        return pow_unit(-1, self).new_quantity(other)
+
     def __div__(self, other):
         return self.__truediv__(other)
-    
+
     def __rdiv__(self, other):
         return self.__rtruediv__(other)
-        
+
     def __pow__(self, other):
         if other == 1:
             return self
         else:
             return pow_unit(other, self)
-        
+
     def __call__(self, x):
         return self.new_quantity(x)
-    
+
     def __eq__(self, other):
         if self is other:
             return True
@@ -128,24 +132,31 @@ class unit(object):
             return self.base == other.base and self.factor == other.factor
         else:
             return False
-    
 
     def __ne__(self, other):
         if isinstance(other, unit):
-            if (isinstance(self, base_unit) and isinstance(other, base_unit)) or \
-                    isinstance(self, nonnumeric_unit) or isinstance(other, nonnumeric_unit):
+            if (
+                (
+                    isinstance(self, base_unit)
+                    and isinstance(other, base_unit)
+                )
+                or (
+                    isinstance(self, nonnumeric_unit)
+                    or isinstance(other, nonnumeric_unit)
+                )
+            ):
                 return NotImplemented
             return self.base != other.base and self.factor != other.factor
         else:
             return True
-   
+
     def __hash__(self):
         return self._hash
-        
+
     @late
     def _hash(self):
         return hash(id(self))
-        
+
     @property
     def dtype(self):
         return None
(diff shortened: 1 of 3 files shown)


rieder:
Using the `|` operator to create quantities is still possible with this PR by the way - but if merged, we should add a deprecation warning to that method.

rieder:
Clearly not yet finished :).

ipelupessy:
how does it do with the tests?

ipelupessy:
its very hard to review with the zillion formatting changes...

rieder:
Lots of errors so far, working on fixing them.
One problem is that the order of constructing units/quantities needs to be taken into account.
Not sure what we can do about that except advising that any unit should be constructed first before assigning it to a value, using parentheses or something.

ipelupessy:
can you give an example of that?

what happens if you keep the old overload available?

rieder:
it's still available, but some constructed units are now suddenly quantities.
Leads to errors like this:
```python
In [4]: (constants.Rydberg_constant * constants.h * constants.c)
   ...:
   ...:
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-4-4f3dad978eb6> in <module>
----> 1 (constants.Rydberg_constant * constants.h * constants.c)
      2

~/Code/amuse/src/amuse/units/quantities.py in __mul__(self, other)
    125     def __mul__(self, other):
    126         other = to_quantity(other)
--> 127         return new_quantity_nonone(self.number * other.number, (self.unit * other.unit).to_simple_form())
    128
    129     __rmul__ = __mul__

~/Code/amuse/src/amuse/units/quantities.py in __mul__(self, other)
    125     def __mul__(self, other):
    126         other = to_quantity(other)
--> 127         return new_quantity_nonone(self.number * other.number, (self.unit * other.unit).to_simple_form())
    128
    129     __rmul__ = __mul__

~/Code/amuse/src/amuse/units/quantities.py in new_quantity_nonone(value, unit)
   1215     :returns: new ScalarQuantity or VectorQuantity object
   1216     """
-> 1217     if not unit.base:
   1218         if isinstance(value, __array_like):
   1219             return numpy.asarray(value) * unit.factor

AttributeError: 'ScalarQuantity' object has no attribute 'base'
```

rieder:
> its very hard to review with the zillion formatting changes... [its not a good sign when github refuses to load the diff]

should be a bit easier now. Main changes are in core.py.

src/amuse/units/derivedsi.py
```diff
@@ -1,7 +1,7 @@
 from amuse.units.si import *
 
-Hz = named('hertz', 'Hz', 1/s)
-MHz = named('megahertz', 'MHz', 1e6*Hz)
+Hz = named('hertz', 'Hz', (1/s).as_unit())
```
ipelupessy, line 3: maybe put the as_unit in named() function?

rieder: good idea!

rieder:
> can you give an example of that?

`c = 299792458.0 * (m*s**-1)` is fine, but `c = 299792458.0 * m*s**-1` is not:
```python
In [7]: c = 299792458.0 * m*s**-1

In [8]: c
Out[8]: quantity<299792458.0 s**-1 m>

In [9]: c.unit
Out[9]: unit<m>

In [10]: c.number
Out[10]: quantity<299792458.0 s**-1>

In [11]: c = 299792458.0 * (m*s**-1)

In [12]: c
Out[12]: quantity<299792458.0 m * s**-1>

In [13]: c.unit
Out[13]: unit<m * s**-1>

In [14]: c.number
Out[14]: 299792458.0
```

src/amuse/units/derivedsi.py
rieder: done in latest commit

ipelupessy:
does the ror still work after this? 

maybe is also good to take stock of all the implications that this change has (issue?):
- old simulation scipts
- documentation
- book 
- sandbox
- downstream (dependend packages ie omuse)
- etc

rieder:
Yes, the ror is untouched and still works.

> maybe is also good to take stock of all the implications that this change has (issue?):
> old simulation scipts
> documentation
> book
> sandbox
> downstream (dependend packages ie omuse)
> etc

Sure. There must not be any remaining issues when this is merged*, so we'll need to test this extensively.

*: other than the creation of new units, which can by design not be done in the old way with this change...

ipelupessy:
quantities has a as_unit and a to_unit - they are slightly different...don't know why.. (i prefer to_unit)

also - there are a bunch of as_units that can be removed and I think in the constants definition we can also remove '* none' units

rieder:
Another issue: 
1.0 * unit will return the unit. This should probably return a quantity instead (in line with previous behaviour and other unit systems)

rieder:
> Another issue:
> 1.0 * unit will return the unit. This should probably return a quantity instead (in line with previous behaviour and other unit systems)

units.kg.__mul__(1.0) does return a quantity. Apparently with 1.0 this isn't even called.

rieder:
ah, of course that would be `__rmul__`...

rieder:
This also fails:
```python
In [31]: 2.0 * 2.0 * kg * m**2
Out[31]: quantity<4.0 m**2 kg>

In [32]: 2.0 * 2.0 * kg**2 * m**2
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-32-7f60638bbeb4> in <module>
----> 1 2.0 * 2.0 * kg**2 * m**2

~/Code/amuse/src/amuse/units/quantities.py in __mul__(self, other)
    125     def __mul__(self, other):
    126         other = to_quantity(other)
--> 127         return new_quantity_nonone(self.number * other.number, (self.unit * other.unit).to_simple_form())
    128
    129     __rmul__ = __mul__

~/Code/amuse/src/amuse/units/quantities.py in new_quantity_nonone(value, unit)
   1214     :returns: new ScalarQuantity or VectorQuantity object
   1215     """
-> 1216     if not unit.base:
   1217         if isinstance(value, __array_like):
   1218             return numpy.asarray(value) * unit.factor

AttributeError: 'ScalarQuantity' object has no attribute 'base'
```

ipelupessy:
hmm, not sure this goes ok....

rieder:
it's really tricky...

rieder:
This is what's happening internally:
```python
In [2]: 1.0 * (kg * m)
mul (kg, m)  # self, other
rmul (kg * m, 1.0)  # self, other
Out[2]: quantity<1.0 kg * m>

In [3]: 1.0 * kg * m
rmul (kg, 1.0)
rmul (m, 1.0)
mul (kg, none)
Out[3]: quantity<1.0 m kg>
```

rieder:
And
```python
In [4]: 1.0 * kg**2 * m
rmul (kg**2, 1.0)
rmul (m, 1.0)
mul (kg**2, none)
rmul (kg**2, 1)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)```

**Comment by rieder** [comment]
In astropy:
```python
In [9]: 1.0 * u.kg**2 * u.m
rmul kg2 1.0
mul m kg2
Out[9]: <Quantity 1.0 kg2 m>```

File: src/amuse/units/core.py [hunk 2]
```diff
@@ -83,12 +83,9 @@ def __truediv__(self, other):
...
 
     def __rmul__(self, other):
-        if other == 1:
-            return self
-        else:
-            if isinstance(other, unit):
-                return factor_unit(other, self)
-            return self.new_quantity(other)
+        if isinstance(other, unit):
+            return factor_unit(other, self)
```
ipelupessy, line 87: this is wrong, because factor units are supposed to construct from a number(like) and a unit..you should only get here with other a number..

ipelupessy:
btw ignore the review, these are just comments

rieder:
comments are useful, I won't ignore them :)

ipelupessy:
would it work if ```1000*m```  resolves to a quantity and m*1000 to a unit?

rieder:
Yes, changing the order seems to work:
```python
In [12]: 3.206361533e-53 * C**3*m**3*J**-2
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-12-545b55826c72> in <module>
----> 1 3.206361533e-53 * C**3*m**3*J**-2

~/Code/amuse/src/amuse/units/quantities.py in __mul__(self, other)
    125     def __mul__(self, other):
    126         other = to_quantity(other)
--> 127         return new_quantity_nonone(self.number * other.number, (self.unit * other.unit).to_simple_form())
    128
    129     __rmul__ = __mul__

~/Code/amuse/src/amuse/units/core.py in to_simple_form(self)
    211                     result = result * base
    212             else:
--> 213                 result =  result * (base ** n)
    214
    215         return result

~/Code/amuse/src/amuse/units/quantities.py in __mul__(self, other)
    125     def __mul__(self, other):
    126         other = to_quantity(other)
--> 127         return new_quantity_nonone(self.number * other.number, (self.unit * other.unit).to_simple_form())
    128
    129     __rmul__ = __mul__

~/Code/amuse/src/amuse/units/quantities.py in new_quantity_nonone(value, unit)
   1214     :returns: new ScalarQuantity or VectorQuantity object
   1215     """
-> 1216     if not unit.base:
   1217         if isinstance(value, __array_like):
   1218             return numpy.asarray(value) * unit.factor

AttributeError: 'ScalarQuantity' object has no attribute 'base'

In [13]: C**3*m**3*J**-2 * 3.206361533e-53
Out[13]: quantity<3.206361533e-53 C**3 * m**3 * J**-2>
```

src/amuse/units/core.py
```diff
@@ -872,10 +872,12 @@ class named_unit(unit):
     >>> (20.0 * (60.0 * si.s)).as_quantity_in(minute)
     quantity<20.0 min>
     """
-    def __init__(self, name, symbol, unit):
+    def __init__(self, name, symbol, local_unit):
         self.name = name
         self.symbol = symbol
-        self.local_unit = unit
+        if not isinstance(local_unit, unit):
+            local_unit = local_unit.as_unit()
```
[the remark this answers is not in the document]
rieder: you already added that I think.

ipelupessy:
did you see the comment on line 87 of core? (__rmul__)

src/amuse/units/core.py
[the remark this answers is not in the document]
ipelupessy: yea, i started a review inadvertedly yesterday, apparently you don't see the comments until a sub,it the review...

src/amuse/units/core.py
rieder: I see

rieder: Replaced it with mul_unit, I think that's correct?

ipelupessy:
also the check for a numerical factor of exactly one should then maybe go to ```__mul__``` or do something with factor_unit, but that is tricky...

rieder:
> did you see the comment on line 87 of core? (**rmul**)

yes, I think it's fixed now. Error is gone at least.

ipelupessy:
can you make a version w/o formatting changes? its very hard to see where it stands now..

ipelupessy:
keep open

PR #696, opened 2020-10-29
Add option to convert quantity to astropy quantity

rieder:
This PR adds an "as_astropy_quantity" method to AMUSE quantities if astropy is available and can be imported.
In case it is not available, it simply doesn't add the method.

### DIFF
--- a/src/amuse/units/quantities.py
+++ b/src/amuse/units/quantities.py
@@ -12,6 +12,12 @@ from amuse.units import core
 from amuse.units.si import none
 from amuse.units.core import zero_unit
 
+try:
+    import astropy.units
+    from amuse.units import units
+    astropy_available = True
+except ImportError:
+    astropy_available = False
 
 """
 """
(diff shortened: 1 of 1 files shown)


rieder:
This would (partially) solve #170.

rieder:
ok to merge?

src/amuse/units/quantities.py
```diff
@@ -12,6 +12,12 @@
 from amuse.units.si import none
 from amuse.units.core import zero_unit
 
+try:
+    import astropy.units
+    from amuse.units import units
+    astropy_available = True
```
ipelupessy, line 18: usually this kind of variable (global in the file is denoted by some combination of capitals and underscores...), so:
__HAS_ASTROPY__=True

ipelupessy:
apart from the small stylistic comment, ok

src/amuse/units/quantities.py
rieder: done

ipelupessy:
one thing I figured just now is that if you don't have astropy installed, you get an atrribute error, maybe its better to raise "install astropy/ astropy not found error" inside the to_astropy function instead of defining anly when astropy is present...

rieder:
sure

PR #706, opened 2020-11-26
add some tools to reduce the tediousness of interface writing

ipelupessy:
for fortran

example use can be found in the omuse qgcm code

opened here for comments

using this you can specify variables and/or parameters in a dict and then generate interface functions on the fortran side, and interface definitions on the python side automatically, steps to do this are the following:

- in the interface.py define the parameters and (grid) variables, like:
```python
parameters={
    "coriolis_f" : dict(short="fnot", dtype="float64", default=0. | units.rad/units.s, description="coriolis frequency", ptype="ro"),
    "coriolis_beta" : dict(short="beta", dtype="float64", default=0. | units.rad/units.s/units.m, description="coriolis beta parameter", ptype="ro"), }
grid_variables={
    "ocean_dynamic_pressure" : dict(pyvar=["pressure"], forvar=["po"], ndim=3, unit=units.m**2/units.s**2, index_ranges=[(1,"nxpo"),(1,"nypo"),(1,"nlo")]), }
```

- add a code generator in the interface.py:
```python
code_generator=FortranCodeGenerator(parameters, grid_variables)
```
- add an include to the interface.f90:
```
include "getters_setters.f90"
```

- add a target to the makefile for this file:
```
getter_setters.f90: interface.py
    	python -c "import interface; print(interface.code_generator_partitioner.generate_getters_setters())" > $@
```

- include an exec in the Interface python definition
```
class CodeInterface(CommonCodeInterface, CodeInterface):

    use_modules=["..."]

    def __init__(self, **keyword_arguments):
        CodeInterface.__init__(self, name_of_the_worker="worker_code", **keyword_arguments)

    exec(code_generator.generate_interface_functions())
```

- and include it in the parameter definitions:
```
    def define_parameters(self, handler):
        code_generator_partitioner.generate_parameter_definitions(handler)
```

it needs some work to also work for the typical nbody codes, but that could be done along the same lines, as it is its mostly handy
for earth science hydro codes with often tens to hundreds of parameters and large number of variables...


ipelupessy:
note it needs a small change in the literature reporting, using stderr instead of stdout

ipelupessy:
I am about to merge this, but @rieder do you think the change in lit. reporting (stderr instead of stdout) is ok? note that an alternative is some option to disable literature reporting...

PR #717, opened 2021-01-06
Update values in SeBa test

rieder:
The SeBa test was giving errors due to masses/velocities not matching the output anymore. I think this is due to updates to SeBa related to massive stars, so I have updated the values. Please check if this is correct (@silviatoonen, @spzwart).

### DIFF
--- a/src/amuse/test/suite/codes_tests/test_seba.py
+++ b/src/amuse/test/suite/codes_tests/test_seba.py
@@ -494,15 +494,15 @@ class TestSeBa(TestWithMPI):
         instance.parameters.supernova_kick_velocity = 0 | units.kms
         instance.commit_parameters()
         print("v_kick=", instance.parameters.supernova_kick_velocity)
-        stars =  Particles(2)
+        stars = Particles(2)
         stars[0].mass = 10.0 | units.MSun
         stars[1].mass = 9 | units.MSun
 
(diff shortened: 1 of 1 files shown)


silviatoonen:
Yes, indeed. We've recently updated the prescription for NS masses, so it makes sense this test fails now and that the numbers need to be updated.

rieder:
So ok to merge @silviatoonen?

rieder:
Changes are in the masses of (massive evolved) stars, but also in the eccentricity and in the natal kick velocity.

ipelupessy:
it looks ok, why is the eccentricity changed to (exactly) 1 where previously it was 0.7 something?

rieder:
> it looks ok, why is the eccentricity changed to (exactly) 1 where previously it was 0.7 something?

I don't know, but I guess this is the result of the evolution of this binary having changed. Perhaps we should choose a different example in this test? @silviatoonen?

silviatoonen:
There is a different kick distribution implemented in SeBa, so the same seed will give rise to a very different kick magnitude and direction, which completely changes the outcome. Best thing to do is pick a different example.

rieder:
After a bit of digging I see that this test was written to test setting/getting the eccentricity of a binary, for which the new value is not a problem.

PR #718, opened 2021-01-12
Change default file format to 'amuse'

rieder:
I never understood why the default file format in Amuse is not 'amuse'. This PR changes that.


rieder:
Any problems with this, @ipelupessy @spzwart?

ipelupessy:
fine with me...

PR #719, opened 2021-01-13
Rename SeBa tests to describe what is tested

rieder:


rieder:
This change works fine with modern testing programs (e.g. pytest) but is incompatible with nose.

PR #723, opened 2021-01-20
Replace instances of tostring() with tobytes()

rieder:
Since the former is deprecated in favour of the latter.


ipelupessy:
so this fixes the numpy deprecation only? (ie all the changed are numpy objects?)

rieder:
I think so - the tests were complaining about all of these and ran successfully afterwards.

rieder:
Ok to merge?

PR #726, opened 2021-02-05
Allow negative weight in FiMap (e.g. velocities)

rieder:
Fix to allow the weight (mass) of a particle in FiMap to be negative (e.g. for using a velocity as weight)


ipelupessy:
I think maybe to be sure mass(index) should be set to zero (in addition to pindex(index)=-1 in map_remove_particle (though clean_array is called before projection so maybe not necessary) - it wouldn't hurt though...for the rest looks ok..

PR #731, opened 2021-02-16
Adding MESA v15140 to AMUSE

rjfarmer:
This is a preliminary pull request to get feedback.

Installable via:
````
DOWNLOAD_CODES=1 make mesa_r15140.code
````

This adds MESA (v15140) as a separate community to AMUSE (#520). This is a nonsdk MESA build so we have to download and install several dependices, but this does mean we can keep using the same gfortran as the rest of AMUSE. Note MESA takes a while to download so for now its setup such that if you have in your shell a path to a local copy, for instance:

````
export AMUSE_LOCAL_COPY_MESA=file:///home/rob/mesa/mesa/mesa-r{version}.zip
````
it will reuse the local copy rather than wasting time re-downloading it.

Interface into amsue can be had by importing:
````
from amuse.community.mesa_r15140.interface import MESA
````

I've tried to keep the interface as similar as possible to old MESA, but there are additions and some removals.

Only the mass, initial\_z, max_model\_number, max\_age, and min_\timestep can be set in ``parameters``, everything else gets deferred to some new routines {get,set}\_{star_job,controls}\_{dble,str,int,logical}, which map to MESA's inlists options, i.e:

````
print(star.get_control_dble('mesh_delta_coeff'))
star.set_control_dble('mesh_delta_coeff',0.8)
````

The reason for this is a) MESA has a lot of options so I'm not writing interfaces for everything and b) there is some annoying data ordering issues, MESA wants options set after a star has been initialized (i.e, after add_particles) while ``parameters`` acts before the star is initialized.

In terms of output there are wrappers for the key stellar proprieties (``star.temperature``, ``star.mass``, etc) which sets units for outputs. But there is also a generic interface:

````
star.get_history('log_center_T')
star.get_profile('h1')
````

Which do not set units, but allows access to all of MESA's history and profile data.

Some things have not been ported, the various set_X_at_zone functions, never made sense in MESA (though the get\_ versions are still there), so they should return an error if called. Its on a todo list to add in an interface to allow altering an existing model to allow importing a structure from another code.

There is currently no documentation, but there are several example scripts in ``examples/``

Other questions?

- The ``patches/makefile_header_non_mesasdk`` currently doesn't use the correct hdf5 linking flags, suggestions?

- We need to check the other link flags are appropriate.

- Error handling is there but most things just return a generic 'something went wrong' so we need to tweak that.

- It needs testing to see what breaks and i missed interfaces, or using bad units etc

Comments? Feedback?

### CODE REVIEW

**Comment by ipelupessy** [comment]
@rjfarmer @rieder just tried this out - one comment I have is that the download script dows some building; I would have the download only do the download and then make the libs; I think this can be put in the makefile themselves rather than in the python script (this may also not propagate the config properly)...

**Comment by ipelupessy** [comment]
furthermore, we may consider to put all the downloaded libraries (except for mesa itself) in the repo?

**Comment by ipelupessy** [comment]
@rjfarmer do you know why the crlibm-1.0beta4 has what seems like a copy of itself in the tar?

**Comment by rjfarmer** [comment]
> @rjfarmer do you know why the crlibm-1.0beta4 has what seems like a copy of itself in the tar?

Nope, probably a bug from when we packaged it up for mesa.

**Comment by rjfarmer** [comment]
Moved all the dependency building out of download.py and into the Makefile.

**Comment by ipelupessy** [comment]
@rjfarmer ok, thanks! I have an issue where I am using a different hdf5 than the one specified in the makefile_header; I think it should be possible to patch the values of the variables HDF5_LIBS and HDF5_FLAGS (available in the AMUSE mesa makefile) into LOAD_HDF5 and INCLUDE_HDF5 respectively...is there anything against this?

**Comment by rjfarmer** [comment]
No I just didn't know what the amuse flags where. There is also blas and lapack flags that need setting, if amuse has an opinion on those as well?

**Comment by ipelupessy** [comment]
no, these are not there (though maybe they should be...hmm...)

**Comment by rjfarmer** [comment]
So what do we have left that needs doing before merging?

There's #747 and #749 anything else we still need to do?

**Comment by rieder** [comment]
on macOS, I am having some trouble at the patching stage:
```
(...)
make -C . patch
mkdir -p /Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140/lib /Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140/include
/Users/srieder/Env/AmuseDev/bin/python ./patch_files.py
checking if quilt is installed ...  no
applying patches to source code
0001-Dont-make-empty-output-directories.patch /Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/patches/0001-Dont-make-empty-output-directories.patch /Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140
patching file star/private/ctrls_io.f90
0001-Fix-astero-builds-when-disabled.patch /Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/patches/0001-Fix-astero-builds-when-disabled.patch /Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140
patching file astero/make/makefile_base
0003-Add-mpi-to-utils.patch /Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/patches/0003-Add-mpi-to-utils.patch /Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140
patching file utils/public/utils_lib.f
0004-Fix-adipls-and-gyre-builds.patch /Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/patches/0004-Fix-adipls-and-gyre-builds.patch /Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140
patching file adipls/adipack.c/adipls/geninh.d.f
0006-Adipls-skip-tests.patch /Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/patches/0006-Adipls-skip-tests.patch /Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140
patch: **** Only garbage was found in the patch input.
Traceback (most recent call last):
  File "/Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/./patch_files.py", line 121, in <module>
    main()
  File "/Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/./patch_files.py", line 107, in main
    apply_patches_using_patch()
  File "/Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/./patch_files.py", line 84, in apply_patches_using_patch
    run_patch(patch, path, os.environ['MESA_DIR'])
  File "/Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/./patch_files.py", line 73, in run_patch
    raise Exception(f"could not apply patch {patchname}")
Exception: could not apply patch 0006-Adipls-skip-tests.patch
make[2]: *** [patch] Error 1
make[1]: *** [download] Error 2
make: *** [/install] Error 2
```

**Comment by rjfarmer** [comment]
I guess it might be stumbling as 0006-Adipls-skip-tests.patch is just meant to make an empty file.

Try 91a152a96 which now makes a non empty file

**Comment by rieder** [comment]
> I guess it might be stumbling as 0006-Adipls-skip-tests.patch is just meant to make an empty file.
> 
> Try [91a152a](https://github.com/amusecode/amuse/commit/91a152a966bdacfb4d9d2a1d2c9323fa6331d30a) which now makes a non empty file

Thanks - solves that problem.
Still having trouble though:

```
LOADER ../tester

3c3
<                                                       B   -1.6215408275458790D-01    1.6391349023885562D+00   -3.8454229229651048D-02   -2.8158487838788577D+00
---
>                                                       B   -1.6215408275458812D-01    1.6391349023885562D+00   -3.8454229229650999D-02   -2.8158487838788577D+00
8c8
<                                                      B2   -1.0851063829787231D+00    9.9999999999999978D-01    1.7975291695264239D+01
---
>                                                      B2   -1.0851063829787233D+00    1.0000000000000000D+00    1.7975291695264243D+01
13c13
<                                               A1_init*X    1.0000000000000000D+00    1.9999999999999998D+00    2.9999999999999996D+00
---
>                                               A1_init*X    1.0000000000000000D+00    2.0000000000000000D+00    2.9999999999999996D+00
15c15
<                                               A2_init*X    1.0999999999999996D+00    2.1000000000000001D+00    3.1000000000000001D+00
---
>                                               A2_init*X    1.1000000000000001D+00    2.1000000000000001D+00    3.1000000000000001D+00

/Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140/mtx/test
TEST FAILED -- compare test_output to tmp.txt


/Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140/mtx
./build_and_test FAILED

make: *** [/lib/libstar.a] Error 1
```

(these seem to be very small roundoff errors, not sure why they cause the test to fail)

**Comment by rjfarmer** [comment]
What gfortran version are you using?

**Comment by rieder** [comment]
10.3.0 (macports version)

**Comment by rjfarmer** [comment]
>(these seem to be very small roundoff errors, not sure why they cause the test to fail)

Normally we sufficiently control everything that is used to compile mesa so that we guarantee getting bit-for-bit the same answers across different machines. So the compile tests looks for exact matches rather than a fuzzy within numerical tolerance match.

On my linux machine with the system compiler (and not the mesa sdk compiler) i get an exact match, so its something weird as usual with a mac.

For now, if you do make an empty file called skip_test in the mesa_r15140/src/mesa-r15140 folder, that will skip the compile time testing and hopefully lets you finish building mesa.

**Comment by rieder** [comment]
That seems to let me progress again a bit further, but only until the next error:
```
/Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140/kap
building kap package.

MAKEDEPF90
COMPILE_CMD ../public/kap_def.f90
COMPILE_CMD ../private/op_def.f90
COMPILE_LEGACY_NOCHECKS ../private/op_load.f
COMPILE_CMD ../private/kap_eval_support.f90
COMPILE_CMD ../private/load_co_kap.f90
COMPILE_CMD ../private/kap_aesopus.f90
COMPILE_CMD ../private/kapcn.f90
COMPILE_CMD ../private/kap_ctrls_io.f90
../private/kap_aesopus.f90:28:7:

   28 |   use hdf5
      |       1
Fatal Error: Cannot open module file 'hdf5.mod' for reading at (1): No such file or directory
compilation terminated.
make[1]: *** [kap_aesopus.o] Error 1
make[1]: *** Waiting for unfinished jobs....

/Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140/kap/make
FAILED


/Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140/kap
./build_and_test FAILED

make: *** [/Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140/lib/libstar.a] Error 1
```

**Comment by ipelupessy** [comment]
you need the hdf5 fortran module, not always included with hdf5 installs...

**Comment by rieder** [comment]
> you need the hdf5 fortran module, not always included with hdf5 installs...

Indeed, the macports version of hdf5 doesn't provide hdf5.mod, even with the +fortran variant...

**Comment by rieder** [comment]
That variant seems provide `libhdf5_fortran.a` instead

**Comment by rieder** [comment]
I think I found some options to get hdf5.mod built, but these are rather obscure and would make installing the prerequisites on AMUSE quite a bit harder. I would much prefer to find a way of not needing hdf5.mod to build mesa, if there is one...

**Comment by rjfarmer** [comment]
The libhdf5_fortran.a file isn't enough (fortran mod files are basically c header files while .a is just a library). 

MESA needs hdf5 as some of its data files are in that format. 

We could switch to building hdf5 during the mesa install?

**Comment by rieder** [comment]
> We could switch to building hdf5 during the mesa install?

Maybe that would be best, yes.

**Comment by rjfarmer** [comment]
Should build hdf5 locally now in 4f5ac41 (also should detect that your on a mac and add the empty file for you)

**Comment by rieder** [comment]
I am trying the example scripts that use MESA with the new version, and not all succeed yet (example(s) below)

**Comment by rieder** [comment]
`merge_two_stars_sph.py` (substituting EVTwin with MESA) results in an error for version 15140:
```
❯ python merge_two_stars_sph.py
Traceback (most recent call last):
  File "/Users/srieder/Code/amuse/examples/textbook/merge_two_stars_sph.py", line 88, in <module>
    p, s = merge_two_stars_sph(**o.__dict__)
  File "/Users/srieder/Code/amuse/examples/textbook/merge_two_stars_sph.py", line 22, in merge_two_stars_sph
    primary_in_sph = return_evolved_star_hydro(Mprim, t_coll, Nsph)
  File "/Users/srieder/Code/amuse/examples/textbook/merge_two_stars_sph.py", line 15, in return_evolved_star_hydro
    star_in_sph = convert_stellar_model_to_SPH(star, Nsph).gas_particles
  File "/Users/srieder/Code/amuse/src/amuse/ext/star_to_sph.py", line 446, in convert_stellar_model_to_SPH
    return converter.result
  File "/Users/srieder/Code/amuse/src/amuse/ext/star_to_sph.py", line 389, in result
    self.retrieve_stellar_structure()
  File "/Users/srieder/Code/amuse/src/amuse/ext/star_to_sph.py", line 79, in retrieve_stellar_structure
    self.composition_profile = self.particle.get_chemical_abundance_profiles(
  File "/Users/srieder/Code/amuse/src/amuse/datamodel/base.py", line 506, in __call__
    return self._function(self.particles, self.particle, *list_arguments, **keyword_arguments)
  File "/Users/srieder/Code/amuse/src/amuse/datamodel/incode_storage.py", line 594, in apply_on_one
    return self.method(index[0], *list_arguments, **keyword_arguments)
  File "/Users/srieder/Code/amuse/src/amuse/community/interface/se.py", line 976, in get_chemical_abundance_profiles
    return self.get_mass_fraction_of_species_at_zone(
  File "/Users/srieder/Code/amuse/src/amuse/support/methods.py", line 168, in __call__
    result = self.method(*list_arguments, **keyword_arguments)
  File "/Users/srieder/Code/amuse/src/amuse/support/methods.py", line 267, in __call__
    return self.method(*list_arguments, **keyword_arguments)
  File "/Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/interface.py", line 824, in get_mass_fraction_of_species_at_zone
    if res['__result'] == 0:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

**Comment by rjfarmer** [comment]
4b3d1506b fixes merge_two_stars_sph

**Comment by rjfarmer** [comment]
b12fd1086 gets stellar_massloss_response.py to run

**Comment by rieder** [comment]
One more issue, encountered running the `helium_star.py` script:
```
> python helium_star.py
Evolving a 12.0 MSun star with MESA until its radius exceeds 100 RSun
Creating helium star, from the inner 350 (out of 1561) shells.
Traceback (most recent call last):
  File "/Users/srieder/Code/amuse/src/amuse/support/core.py", line 518, in memof
    return memof.d[arg]
KeyError: (unit<3.9558027664e+60 * kg**2>, unit<MSun>)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/srieder/Code/amuse/examples/textbook/helium_star.py", line 100, in <module>
    temperatures_helium, luminosities_helium = main()
  File "/Users/srieder/Code/amuse/examples/textbook/helium_star.py", line 40, in main
    helium_star = stellar_evolution.new_particle_from_model(dict(
  File "/Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/interface.py", line 1619, in new_particle_from_model
    cumulative_mass_profile.extend(internal_structure['mass'])
  File "/Users/srieder/Code/amuse/src/amuse/units/quantities.py", line 733, in extend
    self._number = numpy.concatenate((self._number, vector_quantity.value_in(self.unit)))
  File "/Users/srieder/Code/amuse/src/amuse/units/quantities.py", line 231, in value_in
    value_of_unit_in_another_unit = self.unit.value_in(unit)
  File "/Users/srieder/Code/amuse/src/amuse/units/core.py", line 359, in value_in
    return self.conversion_factor_from(unit)
  File "/Users/srieder/Code/amuse/src/amuse/support/core.py", line 520, in memof
    result=f(*arg)
  File "/Users/srieder/Code/amuse/src/amuse/units/core.py", line 302, in conversion_factor_from
    raise IncompatibleUnitsException(x, self)
amuse.units.core.IncompatibleUnitsException: Cannot express 3.9558027664e+60 * kg**2 in MSun, the units do not have the same bases
```

**Comment by rieder** [comment]
`stellar_massloss_response.py` works, but I get very different results from the old mesa (at least when running with the default 1 MSun star): the star seems to experience no noticeable mass loss. I wonder if this is expected behaviour?

**Comment by rieder** [comment]
It doesn't look like the mass change is taken into account at all

**Comment by rjfarmer** [comment]
There's no default wind mass loss in mesa

**Comment by rieder** [comment]
> There's no default wind mass loss in mesa

So this is a change from the old version? We'll need to change the script then...

**Comment by rjfarmer** [comment]
stellar.particles[0].set_control('cool_wind_RGB_scheme','Reimers')
    stellar.particles[0].set_control('Reimers_scaling_factor',0.1)
    stellar.particles[0].set_control('cool_wind_AGB_scheme','Blocker' )
    stellar.particles[0].set_control('Blocker_scaling_factor',0.5)
    stellar.particles[0].set_control('RGB_to_AGB_wind_switch',10**-4)
    
 Would set the wind options back to what you had before.

**Comment by rjfarmer** [comment]
Whats going to be the best way to determine what version of mesa someone is using?

If i do:
````
stellar = MESA()
stellar.version()
````

I get back the git version, so maybe we add an attribute `mesa_version`?

**Comment by rieder** [comment]
@rjfarmer could you pull updates from amuse/main into this branch? One recent commit makes switching between pre/post that commit AMUSE branches a little annoying :).

**Comment by rieder** [comment]
Let's make merging this PR a priority... Getting it merged by the end of January is doable, I think the remaining issues are minor

**Comment by rjfarmer** [comment]
As far as I can tell all the textbook examples that use mesa now run

**Comment by rjfarmer** [comment]
@reider what else do we need to do before merging?

**Comment by rieder** [comment]
I'm testing this today and I'll add any remaining issues here.

**Comment by rieder** [comment]
- src/amuse/community/mesa_next should not be included - please make this a separate PR which can then be merged in time

**Comment by rieder** [comment]
- mesa_r2208 and mesa_r15140 should have separate tests in src/amuse/test/suite/codes_test

**Comment by rieder** [comment]
I still have some trouble with the `helium_star.py` example:
```
❯ python helium_star.py
Evolving a 12.0 MSun star with MESA until its radius exceeds 100 RSun
Creating helium star, from the inner 350 (out of 1561) shells.
Traceback (most recent call last):
  File "/Users/srieder/Code/amuse/src/amuse/support/core.py", line 518, in memof
    return memof.d[arg]
KeyError: (unit<3.9558027664e+60 * kg**2>, unit<MSun>)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/srieder/Code/amuse/examples/textbook/helium_star.py", line 100, in <module>
    temperatures_helium, luminosities_helium = main()
  File "/Users/srieder/Code/amuse/examples/textbook/helium_star.py", line 40, in main
    helium_star = stellar_evolution.new_particle_from_model(dict(
  File "/Users/srieder/Code/amuse/src/amuse/community/mesa_r15140/interface.py", line 1619, in new_particle_from_model
    cumulative_mass_profile.extend(internal_structure['mass'])
  File "/Users/srieder/Code/amuse/src/amuse/units/quantities.py", line 733, in extend
    self._number = numpy.concatenate((self._number, vector_quantity.value_in(self.unit)))
  File "/Users/srieder/Code/amuse/src/amuse/units/quantities.py", line 231, in value_in
    value_of_unit_in_another_unit = self.unit.value_in(unit)
  File "/Users/srieder/Code/amuse/src/amuse/units/core.py", line 359, in value_in
    return self.conversion_factor_from(unit)
  File "/Users/srieder/Code/amuse/src/amuse/support/core.py", line 520, in memof
    result=f(*arg)
  File "/Users/srieder/Code/amuse/src/amuse/units/core.py", line 302, in conversion_factor_from
    raise IncompatibleUnitsException(x, self)
amuse.units.core.IncompatibleUnitsException: Cannot express 3.9558027664e+60 * kg**2 in MSun, the units do not have the same bases
```
Looks like the mass profile is not in fraction of the total mass but already multiplied with the mass?

**Comment by rieder** [comment]
The `helium_star.py` example works now - though the result is quite different from the old MESA (probably can be explained by updates to the code itself?)

**Comment by rieder** [comment]
the `merge_two_stars.py` example still has trouble:
```
❯ python merge_two_stars.py
Traceback (most recent call last):
  File "/Users/srieder/Code/amuse/src/amuse/support/core.py", line 518, in memof
    return memof.d[arg]
KeyError: (unit<none>, unit<Ba>)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/srieder/Code/amuse/src/amuse/support/interface.py", line 622, in convert_arguments
    result[parameter] = quantities.value_in(argument,self.units[index])
  File "/Users/srieder/Code/amuse/src/amuse/units/quantities.py", line 1279, in value_in
    return to_quantity(x).value_in(unit)
  File "/Users/srieder/Code/amuse/src/amuse/units/quantities.py", line 231, in value_in
    value_of_unit_in_another_unit = self.unit.value_in(unit)
  File "/Users/srieder/Code/amuse/src/amuse/units/core.py", line 359, in value_in
    return self.conversion_factor_from(unit)
  File "/Users/srieder/Code/amuse/src/amuse/support/core.py", line 520, in memof
    result=f(*arg)
  File "/Users/srieder/Code/amuse/src/amuse/units/core.py", line 302, in conversion_factor_from
    raise IncompatibleUnitsException(x, self)
amuse.units.core.IncompatibleUnitsException: Cannot express none in Ba, the units do not have the same bases

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/srieder/Code/amuse/examples/textbook/merge_two_stars.py", line 49, in <module>
    radius, rho = merge_two_stars(**o.__dict__)
  File "/Users/srieder/Code/amuse/examples/textbook/merge_two_stars.py", line 20, in merge_two_stars
    stellar.merge_colliding(primary.copy(), secondary.copy(),
  File "/Users/srieder/Code/amuse/src/amuse/community/interface/se.py", line 1052, in merge_colliding
    return merge_colliding_in_stellar_evolution_code(self,
  File "/Users/srieder/Code/amuse/src/amuse/community/interface/se.py", line 1080, in merge_colliding_in_stellar_evolution_code
    col_particle.add_shell(mm1,mass_profile, radius_profile, density_profile,
  File "/Users/srieder/Code/amuse/src/amuse/datamodel/base.py", line 506, in __call__
    return self._function(self.particles, self.particle, *list_arguments, **keyword_arguments)
  File "/Users/srieder/Code/amuse/src/amuse/datamodel/incode_storage.py", line 594, in apply_on_one
    return self.method(index[0], *list_arguments, **keyword_arguments)
  File "/Users/srieder/Code/amuse/src/amuse/support/methods.py", line 167, in __call__
    list_arguments, keyword_arguments = self.convert_arguments(list_arguments, keyword_arguments)
  File "/Users/srieder/Code/amuse/src/amuse/support/methods.py", line 206, in convert_arguments
    return self.definition.convert_arguments(self, list_arguments, keyword_arguments)
  File "/Users/srieder/Code/amuse/src/amuse/support/interface.py", line 624, in convert_arguments
    raise ConvertArgumentsException("error while converting parameter '{0}', error: {1}".format(parameter, ex))
amuse.support.interface.ConvertArgumentsException: error while converting parameter 'pressure', error: Cannot express none in Ba, the units do not have the same bases
```

**Comment by rjfarmer** [comment]
> The `helium_star.py` example works now - though the result is quite different from the old MESA (probably can be explained by updates to the code itself?)

Yes also note that once the model is made, you evolve it for a 1000 steps, which says nothing about how long physically your evolving the model for

**Comment by rjfarmer** [comment]
> * mesa_r2208 and mesa_r15140 should have separate tests in src/amuse/test/suite/codes_test
>       Also: the tests for mesa_r15140 should all pass

Well that's nice to know that exists. Are there any other places with scripts hiding that need updating?

**Comment by rieder** [comment]
I'm just checking any scripts in examples/textbook that contain the word MESA :).

**Comment by rieder** [comment]
> I'm just checking any scripts in examples/textbook that contain the word MESA :).

But there should be no other location with tests, no.

**Comment by rieder** [comment]
A partial run of stellar_isochrone fails to evolve some stars:
```
❯ python stellar_isochrone.py -C MESA15140
Successfully evolved star: m= 0.149133707999 MSun
Successfully evolved star: m= 0.256939557255 MSun
Failed to evolve star: m= 0.100008472493 MSun
Successfully evolved star: m= 0.130558987127 MSun
Successfully evolved star: m= 0.112474265862 MSun
Successfully evolved star: m= 0.107439838338 MSun
```
Not sure why. This is not blocking for me as I had to modify the script to get it to work properly even with the old MESA.

**Comment by rieder** [comment]
Question about the implementation of new MESA vs old MESA:
When running the `stellar_minimal.py` example, I notice that new-MESA is *much* slower than old-MESA in this script.
However, if I change the script to immediately evolve to the final age instead of taking steps of 1 Myr, the difference almost completely disappears. This suggests extra work is being done when taking the smaller timesteps. I wonder if this is necessary or if it could be sped up?

**Comment by rjfarmer** [comment]
> Question about the implementation of new MESA vs old MESA: When running the `stellar_minimal.py` example, I notice that new-MESA is _much_ slower than old-MESA in this script. However, if I change the script to immediately evolve to the final age instead of taking steps of 1 Myr, the difference almost completely disappears. This suggests extra work is being done when taking the smaller timesteps. I wonder if this is necessary or if it could be sped up?

Or that new mesa is taking larger timesteps when allowed to pick its own. Getting through the MS these days normally only needs a few hundred steps (rather than the forced 4700 when taking 1Myr steps).

**Comment by rieder** [comment]
I'm getting a factor 128 in speed difference (3,15 seconds vs 405 seconds), seems a bit too much for just the time step difference?
Also in 'old' MESA this difference is basically nonexistent and it too was forced to make smaller steps...

**Comment by rieder** [comment]
the `merger_stellar_density_profile.py` example has the same problem as the `merge_two_stars.py` example

**Comment by rjfarmer** [comment]
in the codes_test file, how do you update the lines:

````
instance = self.new_instance_of_an_optional_code(MESA)
````

With the mesa version number?

**Comment by rieder** [comment]
I don't think that's possible. Instead could you just import mesa from `amuse.community.mesa_version.interface`

**Comment by rjfarmer** [comment]
``stellar_minimal.py`` had two issues one is that we where taking more substeps than needed. Secondly it seems that 2208 was running with ~600 mesh points (which is a bit low) while the new mesa is using ~3800 mesh points (which is a bit high).

**Comment by rieder** [comment]
Hi @rjfarmer - I'm working on the tests today, and I think I solved a number of issues (will push momentarily).
Some of the tests were really not relevant to the new interface, using outdated methods that I'd rather not see used anymore :). Others are testing unsupported parameters (I have disabled these tests for now, we should discuss if they should be updated or deleted).

One question though, do you know if a year  in MESA is defined as a Julian year?
In AMUSE this is the `units.julianyr` unit, which is not the same as `units.yr` (which is a Gregorian year) - this is probably causing some of the remaining tests to fail.

**Comment by rjfarmer** [comment]
Thanks i was struggling to work out why even a basic stellar model would fail in the test_suite.

MESA's is year is 365.25 days

**Comment by rieder** [comment]
Ok, I've updated the interface to match this

**Comment by rjfarmer** [comment]
Can someone look at the (second) test3 (line 404)? It fails when trying yo access stars[0].luminoisty with:

````
AttributeError: You tried to access attribute 'luminosity' but this attribute is not defined for this set.
````

But there is a getter for luminosity, so it should work?. 

Its possible the currently commented out:

````
        # from_code_to_model = instance.particles.new_channel_to(stars)
        # from_code_to_model.copy()
````

would fix this? But they run into a different issue when uncommented:

````
/data/src/amuse/src/amuse/datamodel/particles.py in get_values_in_store(self, indices, attributes)
   1734                 if len(resultvalue) == 0:
   1735                     if not is_quantity(quantity):
-> 1736                         dtype = quantity.dtype
   1737                         converts[valueindex] = lambda x : x
   1738                         units[valueindex] = None

AttributeError: 'list' object has no attribute 'dtype'
````

**Comment by ipelupessy** [comment]
currently I get some errors with the tests in the build, one I could fix by adapting the test_output (it was a difference in some 10th decimal or so..but this:
```
building rates package.

make[1]: Map '/home/inti/code/amuse/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140/rates/make' wordt binnengegaan
make[1]: Map '/home/inti/code/amuse/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140/rates/make' wordt verlaten
make[1]: Map '/home/inti/code/amuse/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140/rates/make' wordt binnengegaan
make[1]: Map '/home/inti/code/amuse/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140/rates/make' wordt verlaten
make[1]: Map '/home/inti/code/amuse/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140/rates/test/make' wordt binnengegaan
make[1]: Er hoeft niets gedaan te worden voor 'all'.
make[1]: Map '/home/inti/code/amuse/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140/rates/test/make' wordt verlaten

/home/inti/code/amuse/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140/rates/test
FAILED


/home/inti/code/amuse/amuse/src/amuse/community/mesa_r15140/src/mesa-r15140/rates
./build_and_test FAILED
```
I couldn't go around...not sure how to handle this...

**Comment by rieder** [comment]
@rjfarmer I'm having a look now... I think the issue is something is generating a list where it should be an array. Not sure about Inti's issue, it builds fine on my mac...

**Comment by rieder** [comment]
The issue is in lines 1188-1200 of interface.py

**Comment by rieder** [comment]
Commenting these out makes the test work again.

This probably relates to a discussion I was having with Inti today on Slack in some way.

**Comment by ipelupessy** [comment]
I am on linux, and only the adipl has skip_test..

**Comment by ipelupessy** [comment]
the first test error has these differences:
```
3c3
<                                                       B   -1.6215408275458779D-01    1.6391349023885564D+00   -3.8454229229650846D-02   -2.8158487838788582D+00
---
>                                                       B   -1.6215408275458812D-01    1.6391349023885562D+00   -3.8454229229650999D-02   -2.8158487838788577D+00
```
 so that looks ok

**Comment by rjfarmer** [comment]
Whats your gfortran version?

**Comment by ipelupessy** [comment]
so I can just skip_test I guess ;-)

**Comment by ipelupessy** [comment]
gfortran is version 11.2.1

**Comment by rjfarmer** [comment]
[853fcb1](https://github.com/amusecode/amuse/pull/731/commits/853fcb1d4ee352d83b6708bbb9585fb688b1c941) will now skip tests for everyone

**Comment by ipelupessy** [comment]
I think the following fixes the test3 at line 422:
```
        stars=instance.particles.add_particles(stars)
```

**Comment by rieder** [comment]
> I think the following fixes the test3 at line 422:
> 
> ```
>         stars=instance.particles.add_particles(stars)
> ```

Sure, but that's not the real error. The FIXME bit is the real error.
Even `print(instance.particles)` doesn't work currently (but it does when I remove lines 1188-1200 in the interface).

**Comment by ipelupessy** [comment]
its a bug that crops up when return values are non-quantity..

**Comment by rieder** [comment]
I see - so this should be solved either by
- not having these settings as particle attributes
- patching amuse so that it will allow this (if it is a list, convert to array and then check dtype)?

**Comment by ipelupessy** [comment]
its a bug to be fixed in amuse (and maybe small fix in interface): the return type for strings is a list of string
instead of a numpy array of string

there can be two fixes:
in particles.py the ```dtype = quantity.dtype``` can be converted to ```dtype = numpy.array(quantity).dtype```
or in channel.py the corresponding receives of strings can be patched to return numpy.array(strings) instead of the list of strings 'strings'. In this case still the interface needs a slight patch in str2val function to return also numpy arrays

or both can be implemented ;-)

not checked for side effects

**Comment by ipelupessy** [comment]
I am not sure if there are not more places where this bug will crop up (or why we haven't seen it before)

**Comment by ipelupessy** [comment]
as an aside, I think _get_opt must have can_handle_array=True

**Comment by ipelupessy** [comment]
I pushed a fix on main branch; it will need small change to str2val;  to answer my question why we haven't seen it before: its because its shows up in the code of the superset (some extra code needed to convert units)

**Comment by ipelupessy** [comment]
I get an error on:
```
mpiexec -n 1 --oversubscribe pytest-3.8 -sv test_mesa_15140.py::TestMESA::test14
```
if I turn on redirection:
 call star_finish_step
 call star_finish_step
 call star_finish_step
At line 335 of file interface.f90
Fortran runtime error: Bad real number in item 1 of list input

ipelupessy:
also on TestMESAInterface::test7 it fails an assertion because the value "=  2.0000000000000000" (so including equal sign) cannot be converted to float

ipelupessy:
similar thing happens on test8

ipelupessy:
@rjfarmer I think the interface.py can be made somewhat simpler by:
- moving some of the "derived" setters and getters (so the non "legacy function" to the high level interface
- remove most of the define_methods stuff
- define units where necessary in the legacy function definition
I can do this and test, if you agree..

rjfarmer:
@ipelupessy can you try again after deleting and rebuilding the mesa src folder (mesa_r15140/src/mesa*)? I needed to edit the patches applied to mesa to fix those bugs, so mesa itself needs rebuilding

So should the MESAInterface() class just have the python wrappers (things that need legacy) function or remote_function) to the fortran functions? Then any "pure" python functions can sit in MESA()? 

Is there a rule for when things need to be in define_methods? Its always seemed like that if i didn't put something there i'd get an error.

rjfarmer:
Wait skip that the problems still there on test7

ipelupessy:
> So should the MESAInterface() class just have the python wrappers (things that need legacy) function or remote_function) to the fortran functions? Then any "pure" python functions can sit in MESA()?

yea, that is usually the best approach

> Is there a rule for when things need to be in define_methods? Its always seemed like that if i didn't put something there i'd get an error.

define_methods can be replaced by adding a unit = .. to the arguments of the functions addParameters 
if need be you can use the INDEX defined in amuse.community to get the handler.INDEX but even that may not be necessary (handler.NO_UNIT and handler.ERROR_CODE will be done automatically)
the only reason the use a define_methods is if you want a to use kwarg public_name = .. to change the name

rjfarmer:
Okay at this point there should be 25 passing tests with no failures, from the test that run. Note to get them to work you will need to delete mesa_r15140/src/ and run the build again.

ipelupessy:
ok btw, interface.py the stuff at line 1302 is now commented out, is that correct?

ipelupessy:
test pass ok here too

rjfarmer:
They where what was causing things to break earlier with particle sets and things without units. 

For now i just turned them into add_methods rather that getter/setters

ipelupessy:
let me know if you think its ok to merge (doesn't mean you cannot change things later ofcourse..)

rjfarmer:
Maybe we wait until @rieder can test on a mac? Otherwise i think its good to merge

rieder:
All tests pass for me as well. Merging!

PR #732, opened 2021-02-18
AMUSE singularity definition file

ll4strw:
I reckon it could be useful.


ipelupessy:
ok, thanks! we had some previous experiments with singularity recipes, do you rememebr why we did or didn't add a recipe before @rieder ? 

in any case: often the definition file is just called Singularity - is there a reason not to do so?

ll4strw:
The name can be changed as you wish.

ipelupessy:
There is a seperate repository now, so maybe can close?

ipelupessy:
close due to seperate repo

PR #733, opened 2021-02-19
Add mapper, a helper tool for making 2D maps from SPH particles.

rieder:
Makes life easier compared to directly using FiMap.
Plan is to expand this mapper to also use it with Fresco and to add a plotting tool separately.


rieder:
@ipelupessy any thoughts?

ipelupessy:
could be good idea... you may need to add an example for the usage (and maybe some doc string.) though..hard to see how to use it..

ipelupessy:
btw, I think we are using ascii encoding for the source?

rieder:
> could be good idea... you may need to add an example for the usage (and maybe some doc string.) though..hard to see how to use it..

yeah, use case will come in my plotting tool (see https://gist.github.com/rieder/abe1773776a9bb08503e06723d48ca66 for now). I found it better to separate these than have them in one big file.

rieder:
I'm still working on it, syntax will probably change as I add `@property` decorators and such.

rieder:
I simplified it and added a use case example. Still more to be done.

rieder:
I've added (a simplified version of) plot_hydro, which uses mapper.
This seems to work well, but I'm not sure if it is "ready to merge" yet.

rieder:
> hmm, would check your branches and remotes...

I can't see anything wrong. This is part of the right PR and all.

ipelupessy:
ready for merge?

rieder:
Almost. I’ve been using it but I need to add a nice example.

ipelupessy:
suggest merge

rieder:
It's good enough, merging!

PR #735, opened 2021-02-21
Add plotting submodule

rieder:
- Rename `plot.py` to `plot/__init__.py` so it works the same way
- Create a logical place to add plotting modules (TBA)


rieder:
@ipelupessy a very minimal change but I think it would be a good one - with this change we could add plotting scripts in a more logical location than 'ext'.

ipelupessy:
yea, though better still if plot.py -> plot/_plot.py (or plotbase or something) and __init__.py just a from _plot import *
... in any case, these plotting routines are somewhat deprecated...

rieder:
> yea, though better still if plot.py -> plot/_plot.py (or plotbase or something) and **init**.py just a from _plot import *
> ... in any case, these plotting routines are somewhat deprecated...

sure, more work to be done and init should be cleaner by importing from elsewhere (_plot sounds ok). But for backward compatibility, this completely works and allows for incremental updates.

And yes, I guess plot.py is somewhat outdated, but still I can see a use. After all, it's best to avoid requiring the units to be specified multiple times (for plot data and again for axes) if we can...

PR #736, opened 2021-02-23
PR to fix #728

### LINKED ISSUE #728: Add AMUSE version to 'literature printout'
Currently AMUSE prints the literature references, but not the AMUSE release being used.
This should also be printed, either in the form of the version number or as the git hash (or a combination thereof).

Probably this should also contained in the AMUSE config (so that 
```python
import amuse
print(amuse.version)
```
would work)

rieder:
- Automatically generates version number from git tag (+ hash if not on that exact tag) for all packages
- generates `version.py` in package root
- adds AMUSE version to literature printout
- also print version for any community codes (`version` in `amuse.community.(module).version` if it exists, otherwise in `amuse.version`)

Still to do:
- make sure autoconf also knows about this version


ipelupessy:
I will have a look at it..meanwhile, @merijn this one is possibly also relevant for you...

packages/amuse-aarsethzare/setup.py
```diff
@@ -15,7 +13,7 @@
 license_ = "Apache License 2.0"
 url = 'http://www.amusecode.org/'
 install_requires = [
-    'amuse-framework>=%s' % (main_version),
+    'amuse-framework',
```
ipelupessy, line 16: why does the version requirement dissappear?

packages/amuse-aarsethzare/setup.py
```diff
@@ -26,16 +24,23 @@
...
-mapping_from_command_name_to_command_class=setup_commands()
+mapping_from_command_name_to_command_class = setup_commands()
 
 setup(
     name=name,
-    version=version,
+    use_scm_version={
+        "root": "../..",
+        "relative_to": __file__,
+        "write_to": "src/amuse/community/aarsethzare/version.py",
```
ipelupessy, line 41: so this ends up in the package distribution?

rieder: yes

packages/amuse-aarsethzare/setup.py
rieder: Mainly because "main version" is no longer valid with date-based updates. 
We needed this to prevent incompatible versions being installed together.
Maybe we now need to hardcode this instead.

rieder: (and update this whenever an incompatible update is pushed to amuse-framework)

packages/setup_template.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python3
```
ipelupessy, line 1: so you used this to generate all of the setup.py??

rieder: Yes. And then I made changes where needed.

ipelupessy (approved):
looks fine; I will ask more general questions in the pr..

ipelupessy:
so..what does the versioning that is generated look like?

rieder:
It's based on the most recent tag. If that is the current commit (and this should always be the case for a release), the version is identical to the tag (i.e. 2021.2.0).
If not, the version looks something like 2021.2.1.dev9+g02aecc2db.d20210224:
2021.2.1 for the next micro version after the current release
dev9 for 9th commit after this
g02aecc2db for the git commit hash
d20210224 for the date

ipelupessy:
ok..

ipelupessy:
its a nice PR!

rieder:
All community codes will need lines added to their interface like those I already added to Fi and SeBa in order to print the version of that specific module (rather than simply using the parent AMUSE version).

rieder:
New-style output will look something like this:
```

Thank you for using AMUSE!
In this session you have used the modules below. Please cite any relevant articles:


	"AMUSE-Fi (2021.2.1)"
		[1989ApJS...70..419H] ** Hernquist & Katz, 1989
		[2004A&A...422...55P] ** Pelupessy, van der Werf & Icke, 2004
		[2005PhDT........17P] Pelupessy, PhD thesis 2005
		[1997A&A...325..972G] Gerritsen & Icke, 1997


	"AMUSE (2021.2.1.dev14+g441460145.d20210302)"
		[2018araa.book.....P] Portegies Zwart, S. & McMillan, S.L.W., 2018
		[2013CoPhC.183..456P] ** Portegies Zwart, S. et al., 2013
		[2013A&A...557A..84P] ** Pelupessy, F. I. et al., 2013
		[2009NewA...14..369P] Portegies Zwart, S. et al., 2009
```

src/amuse/community/fi/interface.py
```diff
@@ -44,6 +44,13 @@ class FiInterface(
     MODE_NORMAL = 'normal'
     MODE_NORMAL_OPENMP = 'openmp'
     MODE_PERIODIC_BOUNDARIES   = 'periodic'
+    try:
```
ipelupessy, line 47: its better to put something that does this in the parent class (in its __init__)...you may have to use some trickery, but I am pretty sure its possible...then you don't have to go around changing all interface.py....

ipelupessy: in CodeInterface or in the LiteratureReferencesMixIn....

rieder: yes I agree

rieder: I'm not sure how to get the import path right in that case though...

ipelupessy: there is examples of this somewhere in AMUSE....but based on:
```inspect.getabsfile(self.__class__)```

Comment (ipelupessy) [reply src/amuse/community/fi/interface.py:47]: or maybe getsourcefile...

Comment (ipelupessy) [reply src/amuse/community/fi/interface.py:47]: or actually self.__module__ gives you the module...

Comment (rieder) [reply src/amuse/community/fi/interface.py:47]: ok I'll check.
In Python 3.8+ it's much easier... Can just do this then
```python
>>> from importlib.metadata import version
>>> version('amuse-fi')
```

rieder: > or actually `self.__module__` gives you the module...

Ok, I can use that.

ipelupessy: > ok I'll check.
> In Python 3.8+ it's much easier... Can just do this then
> 
> ```python
> >>> from importlib.metadata import version
> >>> version('amuse-fi')
> ```

you could make it so it uses this when available, and just unknown when not...you still need a way to figure out which module you are though..

rieder: Fixed it now.

rieder:
I'm ok to merge it now @ipelupessy

ipelupessy:
ok with me too...

PR #747, opened 2021-04-13
Add configuration option for choosing MESA version

rieder:
- moves mesa to mesa_2208
- adds 'community' class top config.py
- adds "mesa_version" setting that determines which version of MESA is imported with `from amuse.community.mesa.interface import *`


src/amuse/community/mesa/interface.py
```diff
@@ -1,1545 +1,5 @@
...
-                internal_structure.X_Fe[::-1]
-            )
-        tmp_star = datamodel.Particle(key=key)
-        tmp_star.age_tag = current_age
-        return self.imported_stars.add_particle(tmp_star)
-
-
-
-Mesa = MESA
+import amuse
```
ipelupessy, line 1: apparently, git not tracking the move of the interface.py file (and hence any changes) for the interface.py; I don't know if you can fix this (maybe by splitting it up in multiple commits, first moving then generating new file)??

src/amuse/community/mesa/interface.py
```diff
@@ -1,1545 +1,5 @@
...
+if amuse.config.community.mesa_version == "2208":
+    from amuse.community.mesa_2208.interface import *
+elif amuse.config.community.mesa_version == "new":
```
ipelupessy, line 4: "new" or definite version?

src/amuse/config.py
```diff
@@ -108,3 +108,7 @@ class openmp():
     is_enabled = configmk["OPENMP_ENABLED"] == 'yes'
     fcflags = configmk["OPENMP_FCFLAGS"]
     cflags = configmk["OPENMP_CFLAGS"]
+
+class community:
+    mesa_version = "2208"
+    # mesa_version = "new"
```
ipelupessy, line 114: would this work?

```
from amuse.config import community

community.mesa_version="new"

from amuse.community.mesa.interface import MESA
```

src/amuse/community/mesa/interface.py
rieder: yes that's annoying... I'll see what I can do.

src/amuse/config.py
rieder: yes, that's the idea.

src/amuse/community/mesa/interface.py
rieder: new is a placeholder for a better name. 2208 is an exact version but I'm hoping "new" will also accept newer versions than the currently used one, so I don't want to use a number there... I don't like "new" because it won't always be new.

rieder:
Solved in another way

PR #755, opened 2021-04-20
Add Phantom interface

rieder:
PR to add the Phantom interface.
Still need to add tests.


rieder:
@ipelupessy I think it can be merged. The testing could be expanded but I'm realistically not going to find time for that soon...

ipelupessy:
ok, you don't want to make it a package?

rieder:
Also, but it should still be merged for developer installs.

PR #759, opened 2021-04-22
Remove remaining cython bits

rieder:
This was never completed and it won't be...


ipelupessy:
its ok, hopefully we won;t regret this ;)

rieder:
We can always go back and retrieve it.

PR #762, opened 2021-05-11
Fix print statements and deprecated time.clock

MichielsenM:
time.clock doesn't exist in Python 3.8, so I think process_time() (or otherwise perf_counter() ) would be a suitable replacement.


ipelupessy:
I think we should retain them, textbook is text book and seperate; I think the ph4 development thingy may still be used by @stevemcmillan at some point

ipelupessy:
@rieder merge?

rieder:
merged

PR #771, opened 2021-06-24
cleanup install

ipelupessy:
propose to clean everything that is in share:

 - [x] retire build.py
 - [ ] ~move libs to amuse python module tree~ on second thought: any libs by derived packaged are installed in the same dir (otherwise a lot of extra configuration is necessary) so maybe leave as is
 - [ ] ~same for config~ config.mk is also shared between packages (and serves to detect installation type)
 - [ ] ~fix amuse_root_dir and amuse_data_dir (maybe retire the latter)~ for some other pr
 - [ ] ~check the data directory~ this merits its own issue


rieder:
What's the status of this?

ipelupessy:
doesn't have high priority (needs fix)

ipelupessy:
didn't get any further than retiring build.py..

ipelupessy:
the fixes (the definitions see whether they are really necessary etc) of the various (internally) used directories is still worthwhile

ipelupessy:
@rieder I think this one is ok, but it needed a rebuild of the configure script, so if I merge this, be alert to errors here..

ipelupessy:
added: fixed some doc and makefiles

PR #776, opened 2021-07-05
make sure reference is always reported

ipelupessy:


rieder:
Seems good to me. I thought it was already merged. I have no problems if it were merged now.

PR #784, opened 2021-08-10
some convenience to parameters: make code.parameters iterable, allow

ipelupessy:
retrieving the parameter object by name through [] and define the name,
value and description properties

this allows
```python
for p in code.parameters:
  print(p,p.value, p.description)

p=code.parameters["some_name"]
print(p.value)

```


rieder (commented):
This would be quite useful!
Ok to merge for me (I guess you've tested that it works?)

PR #791, opened 2021-09-09
Check which gfortran version is used

### LINKED ISSUE #790: Mesa doesn't compile with gcc < 10 due to unknown option- fallow-invalid-boz
Measa fails to build with compilers older than gcc 10.x due to the option -fallow-invalid-boz hardcoded in src/amuse/community/mesa/Makefile . 
As far as I can see, this option was added because gcc 10.x indeed needs it to be compatible with the code, but older gcc versions don't know the option and crash at that point; editing the Makefile in a local copy makes it again possible to make mesa.code .so, this option should be used only depending on the detected version of the compiler.

rieder:
This fixes #790

### DIFF
--- a/src/amuse/community/mesa/Makefile
+++ b/src/amuse/community/mesa/Makefile
@@ -14,7 +14,12 @@ FORTRAN ?= gfortran
 PYTHON ?= python
 FC      ?= $(FORTRAN)
 MPIFC += $(FCFLAGS)
+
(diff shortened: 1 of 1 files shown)


ipelupessy:
I think this assumes gfortran is used (instead of e.g. intel fortran)..also, there are more codes with this problem (so some general solution might be worthwhile)

Issue #760, opened 2021-04-28
"fi_worker_mp" cannot use more than 200% cpu usage


timroland:
Hello !

**Describe the bug**
When using Fi in "openmp" mode, I cannot get more than 200% CPU usage dedicated to the "fi_worker_mp" process, whatever the environment variable OMP_NUM_THREADS is set to. Openmp does not seem to share correctly the computation load: when running htop, I can see a correct number of threads running (i.e. corresponding to the OMP_NUM_THREADS value), however most of them only use a fraction of CPU power (see screenshot below).

**Expected behavior**
Expect openmp to spread the work load to the number of threads specified by OMP_NUM_THREADS (to scale up my simulations !)

**Logs**
![Fi_openmp](https://user-images.githubusercontent.com/32935374/116440733-10a50680-a851-11eb-8443-af33a6f497ab.png)
In the log file of Fi, the OMP_NUM_THREADS value is also well retrieved.
[Fi.log](https://github.com/amusecode/amuse/files/6393492/Fi.log)

**Environment (please complete the following information):**
I have the same problem on 2 different machines (Amuse is built the same way on both): 
A computing server (80 cores):
 - OS and version: Debian 10
 - Compiler: gcc8.3

and a basic workstation:
 - OS and version: Ubuntu 20.04
 - Compiler: gcc9.3

**Additional context**
Note that on a laptop with Ubuntu 16.04 and gcc7.5 (but only 4 cores), everything works fine. Would it be only a problem of gcc version ?

timroland:
Hi,

We made some investigations out of the AMUSE framework, to test openmp on our server with other codes directly written in fortran, and it seems to work fine, discarding the version issue. We think that the problem would rather reside in the interface between python and fortran. I did not understand how it works exactly but it seems that python launch the processes only on a single core. However, the server is made with only two threads per core, so we would like to use multiple cores to exploit all the threads of the machine (there are 40 cores with 2 threads per core). Would it be possible to overcome this limitation ? 

Thanks again for your help !

ipelupessy:
if its a shared memory machine (e.g. if you do "top" all 80 virtual cpus show up), its a matter of configuring the OpenMP environment, e.g.:
```export OMP_NUM_THREADS=80```
and then start your script, though usually this is not necessary...

timroland:
[...]

PR #796, opened 2021-09-23
add mpi_info_options keyword for code instantiation.

ipelupessy:
proposed solution for #760 and related. This allows setting mpi options for spawned processes.

This allows mpi_info key/value pairs to be passed on to the code. This can be used to for example specify processor binding on a per code basis. you still need to look up the right (and probably mpi implementation specific) mpi_info keywords though.


ipelupessy:
it is possible to specify options in the amuserc under channel in the following form:
```
[channel]
mpi_info_option= key1=value1, key2=value2
```
though values will always be interpreted as strings (maybe that is ok, I don't know..)

ipelupessy:
instance=code(mpi_info_options=dict(bind_to="none"))

PR #801, opened 2021-10-11
retire sei

ipelupessy:
I don't think its ever used, it doesn't have a standard interface ~nor a high level interface,~  its not documented..


ipelupessy:
easily revived

PR #804, opened 2021-11-03
fix MultiprocessingMPIChannel/ fix channel attributes

ipelupessy:
by itself not erribly important, but note that the direct setting of
channel attributes should move to __init__ or start() arguments

### DIFF
--- a/src/amuse/rfi/channel.py
+++ b/src/amuse/rfi/channel.py
@@ -568,6 +568,11 @@ class AbstractMessageChannel(OptionalAttributes):
     @classmethod
     def is_multithreading_supported(cls):
         return True
+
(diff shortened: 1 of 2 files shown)


ipelupessy:
the same fix as for initialize_mpi should/could be done for redirect_stdout_file, redirect_stderr_file and polling_interval_in_milliseconds

ipelupessy:
MultiprocessingMPIChannel itself is not used anywhere atm and could be removed..though it works (it was used for some time to fix testing problems)

the corresponding channel_type ("remote") is slightly misleading, should be something like "mpi_indirect" or "mpi_safe"

ipelupessy:
could be merged as the remaining stuff in the comments can be done later

PR #805, opened 2021-11-05
Add VADER to AMUSE

MJCWilhelm:
Adding VADER (Viscous Accretion Disk Evolutionary Resource, Krumholz & Forbes 2015) to AMUSE, with tests and examples.


ipelupessy:
great! I will have a look asap.. though one question right away: I think the VADER code is now fully contained in the src directory? I think it has a repository on bitbucket - are you using a a different version than that one? Is it an option to download  the VADER code from the repoistory as needed?

ipelupessy:
its not building here because of a missing test.c (?), or at least test.o is a prereq. that can't be build (in src/Makefile, line 12)

MJCWilhelm:
I looked into the issue! Sorry for the delay.

test.cc was a file generated for the amuse interface, which was still in the Makefile but didn't do anything. I now removed it from the Makefile. 

As for the code version, this version has a few edits compared to the original version on bitbucket so it can't be downloaded directly.

rieder:
I would suggest putting the modified version of VADER in a fork of the original repository, and using a download script to get it from there - we do this with most codes now (e.g. PeTar, Seba, ...).

ipelupessy:
thanks! what is the nature and extend of the edits? if they are fixes for the interface and not too many, its also an option to generate patches for them and patch them in during build (but prob. you have considered this...)

MJCWilhelm:
There are just a couple of edits, one fix to a function the interface uses and a change in the parameters of a certain function, order of magnitude 10 lines in total.

I can make a fork for it, but the repository does contain a lot of things that will be redundant for amuse (e.g. a native python interface, its own testing suite, documentation). Would that be a problem?

rieder:
> I can make a fork for it, but the repository does contain a lot of things that will be redundant for amuse (e.g. a native python interface, its own testing suite, documentation). Would that be a problem?

No problem. You could always delete these in your fork if it's too much.

ipelupessy:
most important thing the fork would do is document the steps (in the  commit history) you take to get to the code you are interfacing

ipelupessy:
if you use a fork, also put the changes in a named branch...

MJCWilhelm:
OK, I will set up a fork and have it download from there.

MJCWilhelm:
I set up the fork and an interface branch, but I'm having trouble downloading the archive from bitbucket. It works differently from github, and I can't seem to get a url that I can wget or use in my browser.

I can download the branch from this page:
https://bitbucket.org/MJCWilhelm/vader/downloads/?tab=branches
And then the download url is https://bitbucket.org/MJCWilhelm/vader/get/, with the file MJCWilhelm-vader-0e6357b451e6.tar.gz, but when I combine that into one url I just get 'cannot resolve the cset'. Any idea what I'm doing wrong?

ipelupessy:
think you can download https://bitbucket.org/MJCWilhelm/vader/get/<id>.tar.gz with id branch name or a commit id

MJCWilhelm:
I tried to copy SeBa's download method, but got stuck on Makefile stuff (my old nemesis). I pushed what I have now. If src exists, it just rebuilds and doesn't download, but if src doesn't exist, it complains when it tries to clean. Any ideas?

ipelupessy:
I have a few fixes, but can I remove the current src dir?

MJCWilhelm:
That should be fine, the version in the downloaded repository should be the same. I just renamed it so it was still around but not interfering.

ipelupessy:
so, apperently the current src is different from the src in the repo...so no... @MJCWilhelm ?

ipelupessy:
if i delete src and download the build is not working..

ipelupessy:
(the good news is that if i restore it, it builds)

ipelupessy:
with the above fixes

ipelupessy:
and tests run ok..so please look at the src issue..

MJCWilhelm:
I think this should do it; I forgot that the Makefile in src has to be different from the original one if it is to be compiled for AMUSE. I added it to the vader directory, and it automatically copies it to src. So now I think it should download and build if src is missing?

ipelupessy:
I think download script should do as little as possible...I also removed src and added a line preserving some files that where cleaned up (or make attempted to clean them up..)

ipelupessy:
for me its ok now, though @rieder we sort of decided to always assume DOWNLOAD_CODES=1 ??

ipelupessy:
i made a small change, such that vader_src now is a seperate target; this is convenient when developing because then you can maintain a clone of the vader repo there...

rieder:
> for me its ok now, though @rieder we sort of decided to always assume DOWNLOAD_CODES=1 ??

yes - we need to implement that (by removing DOWNLOAD_CODES references?).

rieder:
but that shouldn't stop us from merging this

ipelupessy:
ready?

MJCWilhelm:
Ready as far as I'm concerned!

PR #817, opened 2022-01-27
adding option to get and set rotation period from SeBa

silviatoonen:


rieder:
It would be good to also add tests for the new methods of course...

rieder:
Ok, the tests are all passing again, so this can be merged (I'll do this).
Still, we should add tests for the new methods too!

PR #819, opened 2022-02-07
Update krome Makefile to work on macOS

rieder:
See https://stackoverflow.com/questions/7573368/in-place-edits-with-sed-on-os-x


rieder:
not tested on non-macOS systems yet (@ipelupessy?)

ipelupessy:
should be ok

Issue #24, opened 2016-01-05
hdf5 cannot write (numpy) unicode strings


ipelupessy:
hdf5 cannot write numpy unicode strings (see https://github.com/ipelupessy/amuse/tree/hdf5_unicode_error), which is a problem because by (i think) commit 3997b47224a29b57807c59c5e8214a11c0cdd5bf  strings returned from code are converted to unicode?

arjenve:
Test is now working, and saving unicode strings should work. Can you confirm?

ipelupessy:
yup, seems to work fine now...

PR #840, opened 2022-03-25
Support HDF5 compression of AMUSE files

rieder:
use with 'compression=True' or 'compression="gzip", compression_opts=compression_level'
(compression level is a number up to 9)

### DIFF
--- a/src/amuse/io/store.py
+++ b/src/amuse/io/store.py
@@ -37,7 +37,7 @@ class HDF5FileFormatProcessor(base.FileFormatProcessor):
     
     provided_formats = ['hdf5', 'amuse']
     
-    def __init__(self, filename = None, set = None, format = None):
+    def __init__(self, filename=None, set=None, format=None):
         base.FileFormatProcessor.__init__(self, filename, set, format)
     
     def load(self):
@@ -49,12 +49,15 @@ class HDF5FileFormatProcessor(base.FileFormatProcessor):
             
     def load_base(self):
         processor = store_v2.StoreHDF(
-                self.filename, 
-                open_for_writing = False, 
-                append_to_file = self.append_to_file or self.allow_writing, 
-                copy_history = self.copy_history,
-                overwrite_file = self.overwrite_file )
-        if not processor.is_correct_version():                
+            self.filename,
+            open_for_writing=False,
+            append_to_file=self.append_to_file or self.allow_writing,
+            copy_history=self.copy_history,
+            overwrite_file=self.overwrite_file,
+            compression=False,
+            compression_opts=None,
+        )
+        if not processor.is_correct_version():
             processor.close()
             processor = store_v1.StoreHDF(
                     self.filename, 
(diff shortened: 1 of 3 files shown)


rieder:
```
In [1]: from amuse.lab import *
   ...: stars = new_plummer_model(1000000)
   ...: write_set_to_file(stars, 'plummer.amuse', compression_opts=None)
   ...: write_set_to_file(stars, 'plummer.amuse_gzip', compression='gzip')
   ...: write_set_to_file(stars, 'plummer.amuse_uncomp', compression=None, compression_opts=None)
   ...: write_set_to_file(stars, 'plummer.amuse_gzip5', compression='gzip', compression_opts=5)
   ...: write_set_to_file(stars, 'plummer.amuse_gzip9', compression='gzip', compression_opts=9)

In [2]: !ls -ltr plummer.amuse*
-rw-r--r-- 1 rieder staff 72164248 25 mrt 16:21 plummer.amuse
-rw-r--r-- 1 rieder staff 54211839 25 mrt 16:21 plummer.amuse_gzip
-rw-r--r-- 1 rieder staff 72012784 25 mrt 16:21 plummer.amuse_uncomp
-rw-r--r-- 1 rieder staff 54211839 25 mrt 16:21 plummer.amuse_gzip5
-rw-r--r-- 1 rieder staff 54211839 25 mrt 16:21 plummer.amuse_gzip9
```

rieder:
compression level doesn't seem to do much, but compression seems to shave off about 50% of the file size at the cost of speed (as could be expected for gzip compression).

ipelupessy:
so does this automatically detect compression in a previous saved file?

ipelupessy:
its already worthwhile if the compression works to shrink arrays with limited information (ie equal mass, coordinates that are the same across columns etc)

rieder:
yes! It's transparent when opening a file.

rieder:
(at least in AMUSE and when using h5py)

ipelupessy:
- I think the PR needs tests
- if its really transparent we may consider making it the default

rieder:
a reason for not making it the default could be that it's definitely slower.
I will add tests.

ipelupessy:
ok, that is good reason; you could also add a store tyoe ('hdf5.gz' or something) as alias for the options as an alternative..

rieder:
I think "compression=True" is a very clear keyword :). I think it suffices.

rieder:
@ipelupessy I put a FIXME in line 105 of store_v2.py, since I thought this was a bit odd:
```
            dtype = numpy.asanyarray(input).dtype
            if dtype.kind == 'U':
                new_dtype = numpy.dtype('S' + dtype.itemsize * 4)
                dataset = group.create_dataset(
                    name,
                    shape=shape,
                    dtype=dtype,  # FIXME new_dtype?
```
I think either this should be `dtype=new_dtype` or we could scrap the new_dtype line.

rieder:
(this is not directly related to this PR, but just something I noticed)

ipelupessy:
hmm, can you see if this is old or added at some point (to maintain compatibility, or for python 2->3??)
there shouldn;t be a reason to have different dtype (hdf5 should accept numpy dtypes)

rieder:
It seems to date back 6 years, related to #24. But clearly it wasn't needed so I'll remove the new_dtype line.

ipelupessy:
yea, the case it catches  should be in a test anyways

rieder:
ok to merge?

ipelupessy:
should be ok

PR #853, opened 2022-05-12
Added acceleration with units to the hermite interface

GijsVermarien:
This change makes the position, vector and acceleration all return amuse quantities in a consistent manner.


ipelupessy:
on second thought, I think that this can be added to gd.py in community/interface??/

GijsVermarien:
Yes I think that would be a good improvement! Even if not all codes in the gd interface have accelerations implemented they might as well return it with correct units if they have it implemented.

PR #861, opened 2022-06-23
Move data dirs

ipelupessy:
to be done:
 - [x] move data dir of evtwin
 - [x] default input data dir in code module directory 
 - [x] move data dirs for other codes
 - [x] implement get_current_working_dir (fortran,c)
 - [x] implement get_Working_directory for java, python (well, java not yet, but placeholders added)
 - [x] check implementation for fortran (ifort and other compilers)
 - [x] change output dir to current_working_dir
 - [ ] ~add (read-only) parameters~ (defer this since not strictly necessary and quite  a work to track down missing parent calls
 - [x] look at mesa_r2208, seems also to have data in src/... and output dir was populated with a file..

(parameters need to be read only because typically setting these is unsafe during run or after init)


ipelupessy:
@rieder do you where the new mesa keeps its data?

ipelupessy:
note that code output now lands in a local __amuse_code_output directory..

ipelupessy:
the only issue now is a very ugly build fix where we need to include the extra function definitions twice (when constructing the worker_code.f90)

rjfarmer:
> @rieder do you where the new mesa keeps its data?

Input micro-physics data is read from mesa-r1540/data 

Though I'm unclear on what your trying to achieve, if you could explain why the changes are happening I can help fix the new mesa.

ipelupessy:
@rjfarmer thanks..we are moving the data for the codes to their respective data directories in the community/code directory..mesa_15140 is not directly affected (because it didn't have its data in the root/data dir..though it would nice if the data is copied or linked to the community/mesa_15140/data directory (will be copied when packaging  in that case..)

rieder:
Looks good. I'll test as much as I currently can (i.e. not the Java codes).

ipelupessy:
@rieder probably can merge this, but should check the resulting packages at some point..

ipelupessy:
packages tests (fi, evtwin)

PR #866, opened 2022-06-29
retire some support stuff

ipelupessy:
cleanup no longer used utilities from the support directory (the distribution support). It no longer worked (well, maybe its fixable) and should be done differently anyway 

maybe also the distribute directory?


ipelupessy:
yea, distribute should retire too:
- its not likely to be working right now, I don't anyone has time to fix it
- the approach taken is not 100% correct, it worked, but but mostly by coincendence (it was supposed to contain all libs, but it didn't, just that most distributions tested had compatible libs
- if revived I would advocate a seperate repo for this kind of tool

rieder:
Agreed...

ipelupessy:
interestingly, the build script was leaking a password (luckily a long shutdown svn repo)

Issue #858, opened 2022-06-12
hop in particle attributes (LagrangianRadii a.o.)


ipelupessy:
some particle attributes are tricky to use because of dependence on Hop..note that the way Hop is used  improves e.g. the determination
of lagrangian radii considerably, so simply removing is maybe not the best option. (although full python implementations of what hop does is not impossible)

Issue #783, opened 2021-08-10
automatically deduce G constant in particle attributes


ipelupessy:
**Is your feature request related to a problem? Please describe.**
a number of particleset attributes for gravitational systems use either constant.G or nbody_system.G (or generic G), e.g. potential.. the functions currently use a default and you have to specify (everywhere I think) if you use nbody_system

**Describe the solution you'd like**
determine automatically as in https://github.com/amusecode/amuse/commit/fcc16fe3163859789498cfbd1eafe42a74dbc9ee

**Additional context**
its a bit of copy paste, maybe move the derive_G function to amuse.units

PR #879, opened 2022-08-16
Auto-set converter in LagrangianRadii

rieder:
Detect if a converter should be used using the base unit of mass.
If so, create the necessary converter using sensible scale units.

### DIFF
--- a/src/amuse/datamodel/particle_attributes.py
+++ b/src/amuse/datamodel/particle_attributes.py
@@ -677,7 +677,7 @@ def mass_segregation_Gini_coefficient(particles, unit_converter=None, density_we
     
     return (mfmnf[1:]+mfmnf[:-1]).sum()/2/(len(mf)-1.)
 
-def LagrangianRadii(stars, unit_converter=None, mf=[0.01,0.02,0.05,0.1,0.2,0.5,0.75,0.9,1],
+def LagrangianRadii(stars, unit_converter="auto", mf=[0.01,0.02,0.05,0.1,0.2,0.5,0.75,0.9,1],
         cm=None, number_of_neighbours=7, reuse_hop=False, hop=HopContainer()):
     """
     Calculate lagrangian radii. Output is radii, mass fraction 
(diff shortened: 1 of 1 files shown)


rieder:
It's a long-standing annoyance that LagrangianRadii needs a converter passed to work. This method would automatically set up a sensible converter when needed.

rieder:
somewhat related to #858

ipelupessy:
I think you can make the converter=None do this (instead of the converter="auto")

rieder:
I wasn't sure of that, since if an explicit "unit_converter=None" is passed we might not want to interfere (even if it will fail).

rieder:
that's why I made "auto" the new default - it's clear enough I think, and setting it to None will force the old (stupid) behaviour.

ipelupessy:
ok, None only makes sense for nbody units

ipelupessy:
I dunno if its also a good time to do #783

src/amuse/datamodel/particle_attributes.py
```diff
@@ -691,6 +691,20 @@ def LagrangianRadii(stars, unit_converter=None, mf=[0.01,0.02,0.05,0.1,0.2,0.5,0
     0.856966667972 length
     """
     import bisect
+    if unit_converter is "auto":
+        # Try to determine the right unit base, using the mass
+        mass_base_unit = stars.mass.unit.base[0][1]
+        if mass_base_unit is units.kg:
+            # Set converter lengths to be something that seems reasonable
+            converter_mass = stars.total_mass() / len(stars)
```
ipelupessy, line 699: I think maybe just the total mass would be better

src/amuse/datamodel/particle_attributes.py
```diff
@@ -691,6 +691,20 @@ def LagrangianRadii(stars, unit_converter=None, mf=[0.01,0.02,0.05,0.1,0.2,0.5,0
...
+            converter_length = (stars.position - stars.center_of_mass()).lengths().mean()
```
ipelupessy, line 700: virial_radius() could also be used..

src/amuse/datamodel/particle_attributes.py
rieder: It probably doesn't matter in any real use case.
My thinking here was that this will make the individual particle mass as close as possible to 1.0.

src/amuse/datamodel/particle_attributes.py
rieder: Yes that makes sense. Do you think that's a better choice? It's a little more expensive but probably not noticeably so.

rieder:
Probably the method to create a converter here could be made available to other methods that require a converter as well...

rieder:
ok to merge? (see open conversations)

ipelupessy:
sure

ipelupessy (approved):

PR #908, opened 2022-11-24
Update literature.py to optionally create tex/bib output

rieder:
This would allow AMUSE to recognise "--bibtex" to automatically create a .tex and .bib file, making it easier to include all relevant references in an article.
Ideally we'd have some string available for each AMUSE package, generated from a bibkey, that can simply be included here.


rieder:
References need to be standardised for this, in the following 
- Each literature line must start with either "ADS:(ADS bibcode)" or "DOI:(DOI code)"
- After this, it doesn't matter, but ideally it would just be the authors' names, year of publishing etc. (not the title)

rieder:
Mostly done. Ideally we should include a bibfile for all codes in AMUSE, probably just one big (not that big) file would do here.

rieder:
I think this works and can be merged.

Issue #856, opened 2022-06-03
not all grid attributes written


ipelupessy:
**Describe the bug**
some grid attributes (of the "hidden" type) like _axes_names or the added derived attributes ("position") are not written with write_set_to_file, hence when reading some 
method (e.g. get_axes_names) do not work

**To Reproduce**
```
g=new_cartesian_grid((5,5), 1)
write_set_to_file(g,"test.amuse","amuse")
g=read_set_from_file("test.amuse","amuse")
print(g.get_axes_names())
```
results in:
```
Exception: do not know how to find axes_names
```

**Expected behavior**
that it works

there are a few possible solutions, like _axes_names moving to collection_attributes, needs a bit of care

ipelupessy:
this needs to be confirmed as fixed or not 100% fixed (think the latter)

Sbte:
I'm running into this as well. I also tried to fix it, but it seems difficult with how everything is set up, and without much knowledge of the underlying code.

PR #923, opened 2023-01-31
grids: Add a method to set axes names.

Sbte:
Workaround for #856.


Sbte:
I would still like to have this merged

ipelupessy:
there will prob. be a more thorough fix for #856, but merging this

PR #928, opened 2023-02-21
Tweak mesa2208 inlist loading

rjfarmer:
The AMUSE_inlist is tweaked to look for a local file called "inlist" (we also patch out the error so it's not an error if this local inlist does not exist). This makes it easier to provide the additional input parameters that amuse does not expose in its interface available to users, without needing to edit the main AMUSE_inlist file.


rieder:
Could you add a parameter switch to the interface to set/unset this (perhaps just setting the filename, defaulting to `None` or `False`)? I don't think we would want AMUSE to behave in a different way depending on the (non) existence of a file by default, but as a parameter it would be explicit and I have no problem with that. It may even be better to fail if the file doesn't exist then.

PR #929, opened 2023-02-27
Expose setting an inlist for mesa 2208

rjfarmer:
This time it's an option set when initialising the Mesa object().


rieder (approved):

rieder:
Looks good, merging.

Issue #172, opened 2017-10-19
Update amuse.units.optparse to use argparse


rieder:
Out of curiosity, why is there a module amuse.units.optparse? 
Is this different from the standard python optparse (which is since Python 2.7 deprecated and replaced by argparse)?
And why is it in amuse.units? Seems a weird place for this...

spzwart:
indeed Steven,
the amuse.units.optparse overloads the units into the standard python package. you can read about it in the amuse book.

Groeten,
Simon

----- Reply message -----

rieder:
No rush though, optparse will not be removed any time soon. https://www.python.org/dev/peps/pep-0389/#deprecation-of-optparse

rieder:
This should be fixed for AMUSE 13 as well.

rieder:
I'll see if I can get it to convert strings. If this works I'll update the examples to that new method and add some deprecation warning to the current optparse method.

rieder:
I've tested @ipelupessy's example above with #934 and it works with that change, so I think that's the way to go.

rieder:
Fixed with #934!

PR #934, opened 2023-03-06
ScalarQuantities now convert strings to floats

rieder:
Possible solution for #172, if it doesn't break anything


rieder:
`units.stellar_type` is probably a case that should be handled differently?

Issue #921, opened 2022-12-26
development install method has changed


rieder:
(discussed with @ipelupessy before on Slack, archiving / continuing here)
See https://setuptools.pypa.io/en/latest/userguide/development_mode.html

Previously, `pip install -e .` would only install AMUSE for pip, it would not build anything. That default behaviour has changed now.
It can (for now) be restored to the old behaviour, by setting `SETUPTOOLS_ENABLE_FEATURES="legacy-editable"`.
We would need to solve this in a more permanent way at some point though.

PR #941, opened 2023-04-03
fix for automatic code building in case of develop (pip -e) install

ipelupessy:
fix for #921

its a bit hacky but seems to work


rieder:
Thanks!
Maybe we need to think about a long-term solution for this though.

rieder (approved):
Seems to work

PR #944, opened 2023-04-10
Remote workers for socket channel

ipelupessy:
This PR allows workers to be started remotely using sockets_channel under the following conditions:
- passwordless ssh access (public key setup)
- amuse is installed
- if an environment needs to be activated with source, specify with remote_env keyword or in amuserc (documented in amuserc.template
- permissions in firewall etc

### DIFF
--- a/src/amuse/rfi/channel.py
+++ b/src/amuse/rfi/channel.py
@@ -1955,7 +1955,7 @@ class SocketChannel(AbstractMessageChannel):
           self.process.stdin.close()
         else:
           logger.debug("starting process with command `%s`, arguments `%s` and environment '%s'", command, arguments, os.environ)
-          print(arguments)
+          # ~ print(arguments)
           self.process = Popen(arguments, executable=command, stdin=PIPE, stdout=None, stderr=None, close_fds=self.close_fds)
 
         logger.debug("waiting for connection from worker")
--- a/src/amuse/rfi/channel.py
+++ b/src/amuse/rfi/channel.py
@@ -913,7 +913,6 @@ class MpiChannel(AbstractMessageChannel):
         self.interpreter_executable = interpreter_executable
                 
         if not legacy_interface_type is None:
-            print(">>", legacy_interface_type)
             self.full_name_of_the_worker = self.get_full_name_of_the_worker(legacy_interface_type)
         else:
             self.full_name_of_the_worker = self.name_of_the_worker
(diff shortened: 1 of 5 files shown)


src/amuse/rfi/channel.py
```diff
@@ -35,8 +35,9 @@
 from amuse.support.options import OptionalAttributes, option, GlobalOptions
 from amuse.support.core import late
 from amuse.support import exceptions
-from amuse.support import get_amuse_root_dir
+from amuse.support import get_amuse_root_dir, get_amuse_directory_root
```
rieder, line 38: these function names are easy to confuse...

src/amuse/rfi/channel.py
```diff
@@ -806,6 +808,8 @@ def split_message(self, call_id, function_id, call_count, dtype_to_arguments, en
         self._communicated_splitted_message = True
         self._merged_results_splitted_message = dtype_to_result
 
+    def makedirs(self,directory):
+        os.makedirs(directory)
```
rieder, line 812: not sure why - could we not just call os.makedirs directly as before?

src/amuse/rfi/gencode.py
```diff
@@ -25,7 +25,7 @@
 from amuse.rfi.tools import create_dir
 from amuse.rfi.tools import create_python_worker
     
-from amuse.support import get_amuse_root_dir    
+from amuse.support import get_amuse_root_dir, get_amuse_directory_root
```
rieder, line 28: easily confused function names...

src/amuse/support/__init__.py
```diff
@@ -26,3 +26,11 @@ def get_amuse_root_dir():
 
 def get_amuse_data_dir():
     return _Defaults().amuse_root_dir
+
+def get_amuse_directory_root():
```
rieder, line 30: easily confused with get_amuse_root_dir, not an ideal name. I get it existed before, but especially when they're in the same module I would prefer clearer names (or at least a docstring that explains the difference)...

src/amuse/rfi/channel.py
ipelupessy: its in the base class to be overridden, ie the default implementation (for local codes) of course needs to call makedirs, but remotely you do something else (distributed or sockets channel)

src/amuse/rfi/channel.py
ipelupessy: yea..in amusifier its called --get-amuse-package-dir, maybe thats better?

ipelupessy: prob consistency is best..

rieder (approved):

PR #952, opened 2023-04-24
Update python-package.yml

rieder:
upgrade pre-installed installed prerequisites


ipelupessy:
looks ok but the package workflows are still not green..

rieder:
yes... the 'get_version' function doesn't seem to work with this action, probably because it's not a proper git repository. Disabling package building for now (didn't work properly anyway)

PR #954, opened 2023-04-26
Update inplace.yml workflow

rieder:
Attempt to fix seemingly random MPI errors when running tests
- Run tests with mpiexec


rieder:
I tried to fix the workflow since it was (it seems randomly) failing some tests.
It seems the tests are now finally working again - at least part of the time (we'll see if this remains true).
The tests are now also a bit more flexible, we can more easily test other environments.
Still some work to be done - e.g. testing other OSes.

PR #956, opened 2023-05-12
Add finer controls for timestepping in mesa

rjfarmer:
MESA has the concept of both a "redo" and a "retry". A redo re-computes the same step with the same dt, but lets you change the physics (say the mass transfer rate). While a retry re-computes a step with a different timestep. This is helpful when evolving multiple stars and you need to keep them in sync.

### DIFF
--- a/src/amuse/community/mesa_r15140/examples/photo_load.py
+++ b/src/amuse/community/mesa_r15140/examples/photo_load.py
@@ -1,0 +1,54 @@
+import matplotlib.pyplot as plt
+import numpy as np
+
+from amuse.units import units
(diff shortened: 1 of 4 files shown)


rieder:
A potential minor issue with this:
if we evolve a (group of) star(s) with MESA in AMUSE and save snapshots in one file (with the timestamp saved), there could be multiple snapshots that share the same timestamp but that have different property values. I wonder if there could be a way to distinguish these. 
Perhaps this is not an issue at all (snapshots/histories are ordered, so just take the last one), in any case not enough to block merging.

rieder (approved):

PR #968, opened 2023-07-06
fix divide by zero warning originating in brokenimf.py

rieder:


rieder:
Not the cleanest solution but I think it's safe

PR #969, opened 2023-07-06
io test fixes

rieder:
fixes a test warning, nothing was really broken


rieder:
Also prevent accidental access to existing files by using tempfile

PR #977, opened 2023-08-30
Updates to SeBa to add functionalities to AMUSE 

silviatoonen:
- the binary type and mass transfer type are now retreavable from within Amuse
- adding option to restart SeBa simulations from within AMUSE, through new_advanced_particle. Need to specify the relative_mass. Optional parameters are stellar type, age, core_mass, CO_coremass and radius


rieder:
One small comment: we should add a test/example for restarting a SeBa simulation using this method.

rieder (approved):

PR #987, opened 2023-09-21
Update howto-install-AMUSE.rst

spzwart:
Added, at the bottom, how to install AMUSE in user space on Snellius (SURF)


rieder:
I've made some changes, now it's using a virtual environment rather than user space - this is safer in case you run multiple projects with slightly different requirements.
Also fixed some typos.

rieder (approved):

rieder:
Also, updated to not build everything but just what is needed - this is much faster (especially MESA will take very long to build otherwise).

PR #988, opened 2023-09-22
Fix bug of adding particles in the middle of simulations for PeTar

lwang-astro:
I have fixed the issue of add new particles in the middle of simulations for PeTar. In the old version, the changeover radii and search radii are not correctly initialized for new particles added in the middle of simulations. If any binaries are added, they are not integrated by using SDAR method, and thus the binary may not be integrated accurately. 

This fix also includes the update of Makefile and download.py. The previous version will download src codes everytime when make is used. Since the download versions are fixed, this is not necessary. In addition, this cause difficulty for recompiling and debugging the codes. 

Two more methods set_output_step/get_output_step are added for better printing of debugging info for PeTar.

### DIFF
--- a/src/amuse/community/petar/Makefile
+++ b/src/amuse/community/petar/Makefile
@@ -40,7 +40,7 @@ LDFLAGS  += -lm $(MUSE_LD_FLAGS)
 
 OBJS = interface.o 
 
-FILELIST=interface.cc interface.py interface.h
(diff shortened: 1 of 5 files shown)


rieder:
Thanks Long! I will run the tests and try to merge this today.

rieder (approved):

PR #994, opened 2023-10-05
remove set_acceleration from all interfaces

### LINKED ISSUE #862: remove set_acceleration from gd.py
propose to remove set_acceleration from the gravitational dynamics interface..its hardly implemented anywhere and I am not sure why it is there (no consistent use case..)

rieder:
fixes #862

### DIFF
--- a/src/amuse/community/adaptb/interface.cc
+++ b/src/amuse/community/adaptb/interface.cc
@@ -443,5 +443,4 @@ int get_kinetic_energy(double* ek){return -2;}
 int get_potential_energy(double* ep){return -2;}
 int get_center_of_mass_velocity(double* vx, double* vy, double* vz){return -2;}
 int get_acceleration(int id, double* ax, double* ay, double* az){return -2;}
-int set_acceleration(int id, double ax, double ay, double az){return -2;}
 
--- a/src/amuse/community/bhtree/interface.cc
+++ b/src/amuse/community/bhtree/interface.cc
@@ -564,12 +564,6 @@ int get_number_of_particles(int *number_of_particles)
   return 0; 
 }
 
-int set_acceleration(int index_of_the_particle, double ax, double ay, double az)
-{
-  return -2;
-}
-
-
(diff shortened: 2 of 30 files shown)


LourensVeen:
I'm not sure if I'm supposed to be reviewing PRs, but why not? Maybe we should discuss that actually, what the process is that we're using.

There's a bit of context missing from the description, but I'm guessing from the comment in community/interface/gd.py that this was physically unsound, and is therefore being removed?

I'm seeing a few more files with `set_acceleration` in them that should probably be updated too:
- `src/amuse/community/mpiamrvac/interface.py`
- `src/amuse/community/mpiamrvac/interface.t`
- `src/amuse/test/suite/core_tests/test_bridge.py`
- `src/amuse/test/suite/codes_tests/test_mpiamrvac.py`
- `doc/tutorial/simplegrav/interface_1.cc`
- `doc/reference/stellar_dynamics_interface_specification.rst`

rieder:
> I'm not sure if I'm supposed to be reviewing PRs, but why not? Maybe we should discuss that actually, what the process is that we're using.
> 
> There's a bit of context missing from the description, but I'm guessing from the comment in community/interface/gd.py that this was physically unsound, and is therefore being removed?

Yes, that's correct. The codes will always overwrite the acceleration, so setting this is misleading.

> I'm seeing a few more files with `set_acceleration` in them that should probably be updated too:
> 
>     * `src/amuse/community/mpiamrvac/interface.py`
> 
>     * `src/amuse/community/mpiamrvac/interface.t`
> 
>     * `src/amuse/test/suite/core_tests/test_bridge.py`
> 
>     * `src/amuse/test/suite/codes_tests/test_mpiamrvac.py`

In these cases it seems to be used in a different way, maybe needs to be addressed but that would be a different PR.

>     * `doc/tutorial/simplegrav/interface_1.cc`
> 
>     * `doc/reference/stellar_dynamics_interface_specification.rst`

Updated.

PR #999, opened 2023-11-01
Delete obsolete files

rieder:


LourensVeen:
Where did the tox.ini information go? Are we not using tox anymore? Pyproject.toml doesn't seem to have a tox section...

rieder:
We replaced Tox with Github actions a while ago.

LourensVeen:
Ah I see, the workflow is calling `pytest` directly. Makes sense.

Issue #1008, opened 2023-11-25
suppress output?


cconroy20:
Is it possible to suppress the text printed to stdout of the form: "In this session you have used the AMUSE modules below...."  It's cluttering up the output information that I'm trying to print to stdout.
thanks
Charlie

rieder:
yes, run your script as `python script.py --no-report-references`.

rieder:
note that we may change this in the future - it is non-documented behaviour and may fail, e.g. when using an argparse/optparse instance.

cconroy20:
Indeed, I have modified the evolution_cluster script to take inputs via argparse, and so this suggestion fails.  When I remove my argparse modifications, then the option "--no-report-references" does work as suggested.  Any ideas for how to get this option to work with additional user-defined input arguments?
thanks

rieder:
Making a 'default' list of arguments to parse is on the to-do list, but has been for a while...
The relevant code is in src/amuse/support/literature.py, line 124.

You could add a dummy argument to your argparser:
`parser.add_argument('--no-report-references', dest='do_not_report_references', action='store_true')`

cconroy20:
that worked - thanks again!

PR #1011, opened 2023-11-27
Add new argparse defaults / deprecate optparse

rieder:
Related to #172 and #1008
- Adds a deprecation message to amuse.units.optparse
- adds amuse.support.argparse, which sets up an ArgumentParser instance with default AMUSE arguments and settings

### DIFF
--- a/src/amuse/support/argparse.py
+++ b/src/amuse/support/argparse.py
@@ -1,0 +1,68 @@
+import argparse
+
+
+def new_argument_parser(
+    **kwargs
+):
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+    )
+    add_amuse_arguments(parser, **kwargs)
+    return parser
+
+
+def add_amuse_arguments(
+    parser,
+    literature=True,
+    units=False,
+):
+    if literature:
+        # Note: at the moment these are dummy arguments, with
+        # amuse.support.literature parsing sys.argv on its own.
+        # They are therefore usable as command line arguments, but the values
+        # here are not used further.
+        parser.add_argument(
+            '--no-report-references',
+            dest='no_report_references',
+            action='store_true',
+            default=False,
+            help="don't report literature references to stdout",
+        )
+        parser.add_argument(
+            '--bibtex',
+            dest='create_bibtex',
+            action='store_true',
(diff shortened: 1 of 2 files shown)


src/amuse/units/optparse.py
```diff
@@ -1,4 +1,8 @@
-
+print(
```
LourensVeen, line 1: Should this use `warnings.warn()` instead of `print()`?

src/amuse/support/argparse.py
```diff
@@ -0,0 +1,68 @@
+import argparse
```
LourensVeen, line 1: It's not clear to me how this replaces the code in `units.optparse`. That code seems to be a helper for doing additional type checking of arbitrary arguments, and for improving formatting of the help text, while this defines (and standardises?) a number of specific arguments that can be passed to AMUSE scripts. So is this a replacement or an addition?

src/amuse/support/argparse.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def new_argument_parser(
+    **kwargs
+):
```
LourensVeen, line 6: If there were a docstring here that could get parsed by the documentation generator, then it would be easier for people to understand how to call this. `**kwargs` isn't the clearest parameter declaration :smile:.

LourensVeen (commented):
Nice to have such a quick response to #1008! I left a few questions, please let me know if I can help.

src/amuse/units/optparse.py
rieder: Yep. But couldn't get that to work from the train. Feel free to fix.

src/amuse/support/argparse.py
rieder: #172 (I think) details how optparse is obsolete- attaching a unit is now possible by specifying type=units.(unit). This PR explicitly makes optparse deprecated, encouraging the use of argparse and making our literature arguments more accessible. Also, the new_argument_parser here will show the default settings, which was missing before.

src/amuse/support/argparse.py
rieder: Absolutely. 
Just felt I had to split these two functions in order to make the arguments available without the need for a new parser. 
But that could probably be solved in a more practical way.

PR #1016, opened 2023-12-06
Update patch_files.py for mpiamrvac

rieder:


LourensVeen:
FWIW, if you'd called the commit `Reformat patch_files.py for mpiamrvac` it would have been immediately clear what this does. Makes it easier for other developers to follow along, or for people rooting around the git history later.

rieder:
that's part of the update, but there was a bugfix in there too (line 43, convert to str).
I admit, better to separate those...

PR #1019, opened 2023-12-11
Update rebound interface to Rebound release version 4.0.1

rieder:
The Rebound version currently in AMUSE is quite old - this updates it to the latest release.
The current tests for the rebound interface still work, but should probably be extended.

### DIFF
--- a/src/amuse/community/rebound/Makefile
+++ b/src/amuse/community/rebound/Makefile
@@ -20,15 +20,37 @@ OBJS = interface.o
 
 CODELIB = src/librebound.a
 CODEDIR = src/rebound
-LIB_OBJ = $(CODEDIR)/src/rebound.o $(CODEDIR)/src/tree.o $(CODEDIR)/src/particle.o $(CODEDIR)/src/gravity.o \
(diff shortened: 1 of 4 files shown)


LourensVeen (approved):
LGTM, nice!

rieder:
It needs more extensive testing. Some features seem to work but on closer inspection require some settings to be changed, not all of which are available through the interface.

LourensVeen:
Hm, so I was early. Should there be some kind of norm for this? When is a code integrated well enough that the PR adding it can be merged? That would then set some standards for reviews. All assuming that we're rich enough to be able to afford these kinds of things of course...

rieder:
I think in principle it can be merged (since it passes all current tests).

However, since the new version of Rebound adds features, these need to be tested as well.
So ideally we want some covering check as well, to make sure the tests don't become meaningless.

LourensVeen:
That makes sense.

rieder:
Rebound should probably have a second worker built - one with, and one without OpenMP support.
For the `whfast` integrator (which is in turn used by some other integrators), behaviour is different depending on whether OpenMP is used. 
In the OpenMP case, it needs a different coordinate system (which is not available through the AMUSE interface yet, and I need to check whether it would make sense to have this available at all).

PR #1022, opened 2023-12-14
Move code tests to code dirs (attempt 2)

rieder:
The idea is that the tests for a specific community code (which is available as a separate package) should be included with that code, and not in a centralised place.


rieder:
If this PR is merged, the tests for a code will be available under `src/amuse/community/{codename}/test_{code}.py`.
Maybe it should be even simpler, `src/amuse/community/{codename}/test.py`?

LourensVeen:
Assuming the tests will all be in Python, then making a `src/amuse/community/{codename}/tests/` directory with one or more `test_<something.py` files would be the standard way to do it. Pytest will auto-discover this, and we could set it up so that you can run pytest from the root to run all tests, or from a community code subdirectory to only run its tests.

LourensVeen:
Hmm, I think this is a good idea, and it's tempting to do it while working on the new build system. I will want to test things anyway to reduce the chances of breaking anything, and since I'm setting up a new system to build and install, I might as well add running tests to it. If it's just moving the files...

LourensVeen:
This is now being done as part of the move to the new build system. There seem to be a bunch of unrelated updates in the final commit related to the build system and the CI, and whatever that does will have to be done there as well anyway. So I propose we close this?

PR #1027, opened 2024-01-19
refactor support/interface.py with black

rieder:


LourensVeen:
Looks black! :smile:

I also agree that 88 characters is much better than 80. I've recently been switching my other code bases and it's really noticeable.

rieder:
> Looks black! 😄
> 
> I also agree that 88 characters is much better than 80. I've recently been switching my other code bases and it's really noticeable.

Black has 88 as the default, I switched it to 80 in the second commit since I noticed errors with this in my editor.
But I think you're right, 88 looks better.

LourensVeen:
Oh, I misread, I thought you'd gone from 80 to 88!

I think the rationale behind 88 is that most code in larger programs is in classes, and so you always lose 8 characters to the indentation. In my experience, those 8 extra characters surprisingly often mean that something just fits on a line as opposed to having to be split, and it makes everything looks better.

I bet your editor has a line length setting somewhere :smile:

rieder:
Yeah, I can see why 88 makes more sense. I'll adjust the value in my editor and revert the second commit.

PR #1028, opened 2024-01-24
Let hermite end at exactly the requested time by default

rieder:
set default value of hermite.parameters.end_time_accuracy_factor from 0.5 to 0.0


rieder:
Pro: Hermite will stop at exactly the expected time.
Con: Hermite will give different results depending on how many evolve steps are taken.

The former seems more important as the default.

PR #1039, opened 2024-03-14
quantities: Make mean() more generic.

Sbte:
I ran into this while trying to pass the `where` argument to `numpy.mean()`.


LourensVeen:
Looks fine to me. It does extend the AMUSE API to match all of the numpy API (and it means we'll evolve with numpy if they change) but we're doing that already for `median()`, so I'd say that's alright.

Is there any particular usage you have that you want to add a test for?

Sbte:
Thanks for the review Lourens. I added a test. I'm not sure why the tests are numbered instead of using meaningful names like `test_mean`, but I followed the same pattern.

rieder (approved):

PR #1045, opened 2024-04-23
Update ci.yml

rieder:


.github/workflows/lint.yml
```diff
@@ -15,17 +15,18 @@ jobs:
...
 
       - name: Install Python dependencies
-        run: pip install flake8
+        run: pip install flake8 black
 
       - name: Run linters
         uses: wearerequired/lint-action@v2
         with:
+          black: true
           flake8: true
```
LourensVeen, line 32: Should we only call black then, and not flake8?

LourensVeen (commented):
Nice, but we should probably pick one tool and not run two, as they may end up fighting each other. Also, possibly [ruff](https://docs.astral.sh/ruff/) should be under consideration too?

.github/workflows/lint.yml
rieder: This is just for checking, not fixing. So I don't mind running both. For fixing syntax we should make a choice.

rieder:
> Nice, but we should probably pick one tool and not run two, as they may end up fighting each other. Also, possibly [ruff](https://docs.astral.sh/ruff/) should be under consideration too?

yes we should decide which tool(s) we would want. black and flake8 don't do exactly the same thing, so for checks we may want to use both. Or ruff, I'll have a look at that too.

rieder:
@LourensVeen I'm not sure how to test the auto-fix part of this, do you have any ideas for that?

rieder:
I've tested it on a different repository now. The linter will push a fix to the branch in the PR, which seems fine to me.

rieder:
I'm just not sure yet if this will fix only the files affected by the PR (which should be how this works) or if it checks all other files too (which I really really don't want). Need to be sure of that before merging.

PR #1058, opened 2024-05-29
Added collision handler for Huayno OK algorithm

ErwanH29:
Integrated the collision detection algorithm from the various SHARED_COLLISION Huayno integrators into the OK integrator. 
Note: This collision checker does not detect collisions between two massless particles.


ErwanH29:
I made a new file which hosts the collision detection function and with which both evolve_shared_collisions.c and evolve_ok.c call. The algorithm neglects collisions between two massless particles to save computing time at the detriment that perhaps this isn't an assumption that some users would want. If that is a bother I can rewrite the code to include it.

I hope the changes are fine. I tested collisions using either integrators and I don't see a change in performance.

rieder:
This PR needs at least a test to show that this collision detection actually works before it can be merged.

PR #1059, opened 2024-06-14
split off astronomical constants from units

rieder:
Astronomical constants have changed, this is part of an effort to make these easier updateable.
units.py should not contain changeable values anymore.

### DIFF
--- a/src/amuse/units/astronomical_constants.py
+++ b/src/amuse/units/astronomical_constants.py
@@ -1,0 +1,19 @@
+"""
+Series of astronomical constants, which are in turn used by the units of the
+same names.
+"""
+
+import numpy as np
+from amuse.units.si import m, kg
+from amuse.units.derivedsi import W, km
+
+au = 149597870691.0 | m
+parsec = au / np.tan(np.pi / (180 * 60 * 60))
+lightyear = 9460730472580.8 | km
+Lsun = 3.839e26 | W
+Msun = 1.98892e30 | kg
+Rsun = 6.955e8 | m
+Mjupiter = 1.8987e27 | kg
+Rjupiter = 71492.0 | km
+Mearth = 5.9722e24 | kg
+Rearth = 6371.0088 | km  # IUGG mean radius
--- a/src/amuse/units/units.py
+++ b/src/amuse/units/units.py
@@ -4,6 +4,7 @@ Units supported in AMUSE
 
 import numpy
 from . import constants
+from . import astronomical_constants
 from . import quantities
 
 # The two imports below are to explicitly expose everything directly used in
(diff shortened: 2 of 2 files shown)


LourensVeen (dismissed):
Looks good! Metallicity also sounds quite astronomical to me, but it doesn't have an associated constant. Maybe we should have a look at splitting up the units, probably in conjunction with reintegrating OMUSE...

rieder:
There is no constant for metallicity. The closest would be 'solar metallicity', but even for that there isn't a generally accepted constant value.
Splitting up the units might be a good idea.

LourensVeen:
Yeah, and it's a fraction anyway, right? It just occurred to me that metallicity was at the bottom of the file, amidst degrees and grams and centimeters which are not discipline-specific. Anyway, something for later. Will you merge this?

rieder:
There's a remaining issue: names are used both for the constants and for the unit based on the constant.
So we need to rethink this or be very careful with namespaces.
Something that does `from amuse.units.constants import *` will now cause problems - as is the case in one of the tests (test_pickle.py). Then again of course, this is /exactly/ the reason why you should *never* import *...

rieder:
(Of course I'm violating that "rule" myself in a commit only minutes after posting this...)

src/amuse/test/suite/core_tests/test_pickle.py
```diff
@@ -1,37 +1,30 @@
...
 
 class TestPicklingOfUnitsAndQuantities(amusetest.TestCase):
 
     def test1(self):
-        km = 1000 * m
-        self.assertEqual(1000, km.value_in(m))
-        pickled_km = pickle.dumps(km)
-        unpickled_km = pickle.loads(pickled_km)
-        self.assertEqual(1000, unpickled_km.value_in(m))
+        kilometer = 1000 * m
```
rieder, line 19: renaming it here since 'km' is already defined... could probably just as easily use the pre-defined 'km' unit, but maybe there was a reason not to?

src/amuse/units/amuse_2010/astronomical_constants.py
```diff
@@ -0,0 +1,19 @@
...
+same names.
+"""
+
+import numpy as np
+from amuse.units.si import m, kg
+from amuse.units.derivedsi import W, km
+
+
+au = 149597870691.0 | m
+parsec = au / np.tan(np.pi / (180 * 60 * 60))
```
rieder, line 12: Note that both 'au' and 'parsec' are also defined as units... Be careful, this may clash!
For Lsun etc below, note that here they only have one capital letter, the unit has LSun etc... This is to be more in line with Astropy etc, which only have one capital letter. Still, it may be better to be consistent within AMUSE?

src/amuse/units/nist.py
```diff
@@ -45,157 +50,213 @@ def get_table_from_url(self):
...
+        mydiff = difflib.unified_diff(
+            self.nist_table.splitlines(1), self.local_table.splitlines(1)
+        )
         for i in list(mydiff):
             print(i)
-            
+
     def get_translator(self):
-        f = open(os.path.join(self.directory, 'translator.txt'), 'r')     
+        f = open(os.path.join(self.directory, "translator.txt"), "r")
```
LourensVeen, line 83: Please keep real changes and formatting changes from black in separate commits, can't see the forest for the trees here...

src/amuse/units/amuse_2010/astronomical_constants.py
rieder: normally it wouldn't be a problem, since units would be used as `units.(unitname)` rather than being directly imported. But I don't exactly like it.

LourensVeen: Are constants that can also be units actually ever used as anything else than units?

rieder: Probably not. Except for making the units.

rieder: So... We could in principle make the names here (way) more descriptive, e.g.:
```
distance_parsec = ...
distance_au = ...
luminosity_sun = ...
radius_sun = ...
```
That would solve this issue. It's very verbose but since you're always using the unit form that should not be a problem.

LourensVeen: Well, they could still have the same names actually, as long as no one directly uses the `amuse.units.amuse_2010.astronomical_constants` module. It should only be used in the modules that define the units then, and we could even name it `_astronomical_constants` to mark it as private.

Then in the modules where the units are defined, we could `import amuse.units.amuse_2010.astronomical_constants as ac` and write `ac.au` and the like everywhere to clearly point out where they're coming from and that these are the constants and not the units. I think that's what I'd do.

PR #1066, opened 2024-08-12
Update Phantom interface

rieder:
Large update to the Phantom interface: fixing bugs, adding features and updating the code version used


rieder:
Need to move this to the new build system, but apart from that I think this is done.

PR #1067, opened 2024-08-15
Cleanup Fi interface syntax

rieder:
- syntax updated with Black
- use explicit imports

### DIFF
--- a/src/amuse/community/fi/interface.py
+++ b/src/amuse/community/fi/interface.py
@@ -1,1480 +1,1605 @@
-import os
-import numpy
-from amuse.community.interface.gd import GravitationalDynamicsInterface
-from amuse.community.interface.gd import GravitationalDynamics
-from amuse.community.interface.gd import GravityFieldInterface
-from amuse.community.interface.gd import GravityFieldCode
-from amuse.community import *
-from amuse.support.options import option
+"""
+Interface for the Fi smoothed-particle hydrodynamics code.
+"""
+from amuse.community import (
+    CodeInterface, LiteratureReferencesMixIn,
+    StoppingConditionInterface, StoppingConditions,
+    CodeWithDataDirectories,
+    legacy_function, LegacyFunctionSpecification
+)
+from amuse.community.interface.gd import (
+    GravitationalDynamicsInterface, GravitationalDynamics,
+    GravityFieldInterface, GravityFieldCode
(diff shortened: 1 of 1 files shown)


src/amuse/community/fi/interface.py
```diff
@@ -1,2414 +1,2599 @@
...
-        function.addParameter('id', dtype='i', direction=function.IN)
-        for x in ['mass','x','y','z','vx','vy','vz','u',]:
-            function.addParameter(x, dtype='d', direction=function.IN)
-        function.addParameter('h_smooth', dtype='d', direction=function.IN, default = 0)
-        function.result_type = 'i'
-        return function
-    
-    @legacy_function    
+        function.addParameter("id", dtype="i", direction=function.IN)
+        for x in [
```
LourensVeen, line 176: Ouch! Oh well, it is what it is with autoformatters...

src/amuse/community/fi/interface.py
```diff
@@ -2608,1484 +2769,1433 @@ def define_methods(self, handler):
...
-            (handler.NO_UNIT,),
-            (handler.ERROR_CODE,)
+            (
+                handler.NO_UNIT,
+                handler.ERROR_CODE,
+            ),
         )
-        
+
+        handler.add_method("set_halofile", (handler.NO_UNIT,), (handler.ERROR_CODE,))
```
LourensVeen, line 3309: Now this is inconsistent with everything else... Again, ah well...

LourensVeen (approved):
Looks good, I like the comment at the top too.

src/amuse/community/fi/interface.py
rieder: It makes sense - the autoformatter would add spaces, and with the spaces I guess it exceeds the max line length (which defaults to 80 rather than 88 - may have to change that).

src/amuse/community/fi/interface.py
rieder: Yes, not sure why.

src/amuse/community/fi/interface.py
LourensVeen: 88 is surprisingly much better than 80, so let's change it indeed. And maybe do it right away and reformat this and then merge it, so that we don't have to do all this again later.

PR #1069, opened 2024-09-03
Allow named units to have a LaTeX representation (WIP)

rieder:


rieder:
Meant to allow for nice LaTeX representations like R_{\odot} instead of RSun

LourensVeen:
Sounds like a good idea. The one thing that comes to mind is that these symbols are probably in Unicode, and that we could do that. But I'm not sure how well Unicode works in LaTeX, although it will probably work in plotting software.

Maybe something to test, it's nicer and more modern. On the other hand, LaTeX is still very widely used in science, so if Unicode doesn't solve the problem then it'll be good to have for sure.

rieder:
Ready to merge I think.

LourensVeen (approved):
That's nice and simple, doesn't hurt anything, and can be expanded over time with more LaTeX as people want it or if someone gets bored :smile:.

PR #1078, opened 2024-10-09
Work to refactor amuse.rfi

rieder:


src/amuse/rfi/channel.py
```diff
@@ -44,9 +44,11 @@
 from . import async_request
```
LourensVeen, line 34: This file is such a disaster. But at least it's now a nicely formatted disaster. Great!

src/amuse/rfi/import_module.py
```diff
@@ -68,64 +75,71 @@ def import_unique(modulename):
...
 def callback(info, size, data):
     # simple search
     print("CLEANUP:", info.contents.dlpi_name)
     count[0] += 1
     return 0
-  
+
+
 def cleanup_module(mod):
-    #print "CLEANUP!!"
```
LourensVeen, line 109: Maybe this Python 2 comment could just go?

LourensVeen (commented):
Looks good overall but I've left some questions and a proposal for one more fix.

PR #1081, opened 2024-10-21
Update plot_hydro_density_distributions.py: print the right info

rieder:


examples/textbook/plot_hydro_density_distributions.py
```diff
@@ -13,7 +13,7 @@ def stellar_model(N, M, t=0.0|units.Myr):
     se_star = stellar_evolution.particles.add_particle(star)
     print("Evolving", star.mass, "to t=", t.in_(units.Myr))
     stellar_evolution.evolve_model(t)
-    print("Stellar type:", stellar_evolution.particles.stellar_type)
+    print("Stellar type:", se_star.stellar_type)
```
rieder, line 16: This will print "Main Sequence Star" rather than "[1] stellar_type"...

PR #1083, opened 2024-10-23
Fix amusifier Python 3.13 compatibility

### LINKED ISSUE #1082: Amusifier breaks on Python 3.13
**Describe the bug**
Amusifier crashes on Python 3.13, because `None` has a docstring in 3.13 and `base.py:531` assumes that that is not the case. It should really be checking for `is None` before doing anything with `function`.

There may be other places where things fail on Python 3.13, this should be tested and fixed as needed as well.

Thanks to Niels from this year's AMUSE course for reporting!

LourensVeen:
Here are some small fixes to make the amusifier not crash on Python 3.13. I've tested this manually on 3.13 and also on 3.8 to check for regressions, and it works on both.

To explain:

In attributes.py, `dict` is the namespace of the class, so it contains the parameter declarations in the classes but also a bunch of variables made by Python. The latter all have dunder names, the former don't, but instead of checking the name of the key, we check the type of the value. Python 3.13 added a new variable that happens to have an empty tuple as its value, so we crash.

In `base.py`, `function` is `None` (by design, apparently) and `None` used to have a `__doc__` that was `None`, but in Python 3.13 it's a string describing the `None` object. So it's no longer the case that any object that has a `__doc__` also has a `__name__`, because now `None` has the former but not the latter. That was questionable anyway. So there's an extra check for `None` here.

Fixes #1082

### DIFF
--- a/src/amuse/datamodel/attributes.py
+++ b/src/amuse/datamodel/attributes.py
@@ -17,7 +17,7 @@ class DomainMetaclass(type):
     def __new__(metaclass, name, bases, dict):
         replacement_dictionary = {}
         for key, value in dict.items():
-            if isinstance(value, tuple):
+            if not key.startswith('__') and isinstance(value, tuple):
                 default_value, description = value
                 replacement_dictionary[key] = AttributeDefinition(
                     key, description, default_value.unit, default_value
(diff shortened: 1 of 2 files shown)


rieder (approved):
Looks OK and test-building `hermite` works with this patch on Py3.13

PR #1085, opened 2024-10-30
Update MANIFEST.in

rieder:


LourensVeen (approved):
Looks good, no point in pruning something that doesn't exist.

PR #1086, opened 2024-10-30
Issue 940: update mpreal.h

### LINKED ISSUE #940: update necessary for mpreal.h (adaptb, brutus)
I am not sure yet why (whether its the compiler or mpfr version that requires it) but brutus and adaptb 
need the latest mpreal.h (https://github.com/advanpix/mpreal/blob/master/mpreal.h)

not sure it will work on older compiler /mpfr...

LourensVeen:
This fixes #940 by replacing mpreal.h for both Brutus and Adaptb, updating Brutus to compile with newer compilers (`using namespace std;` is a bad idea because the namespace may gain new members over time that conflict with your code, as happened here in C++17), and fixes a C/C++ symbol confusion problem when linking Brutus.

I tried it out with the simple gravity simulation from `examples/textbook/gravity_minimal.py` and it works, except that adaptb doesn't implement some of the methods, but that is what it is given that it's deprecated anyway. At least they compile and run again.

### DIFF
--- a/src/amuse/community/adaptb/src/mpreal.h
+++ b/src/amuse/community/adaptb/src/mpreal.h
@@ -1,34 +1,35 @@
 /*
-    MPFR C++: Multi-precision floating point number class for C++. 
+    MPFR C++: Multi-precision floating point number class for C++.
     Based on MPFR library:    http://mpfr.org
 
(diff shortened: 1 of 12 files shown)


rieder (approved):
looks good

PR #1093, opened 2024-11-26
Cleanup ic.molecular_cloud

rieder:
- Slightly better syntax
- More consistent keyword naming, with a warning for when the old style is used 
- No more default number of particles, must be set explicitly


rieder:
Can be merged I think

rieder:
@LourensVeen could you approve this? I'd like it merged to have more consistency in the book examples.

rieder:
@LourensVeen please approve if you have no objections :)

PR #1095, opened 2024-12-03
Fix uninitialised counter variable

### LINKED ISSUE #1094: Huayno crashes with "ENDRUN at ../src/evolve_sf.c:327 partition error"
**Describe the bug**
Huayno crashes with "ENDRUN at ../src/evolve_sf.c:327 partition error".

This happens because `i` isn't initialised in that routine. It seems that most of the time it ends up being 0 anyway, but on at least one occasion (macOS, clang) it seems like it wasn't 0. At any rate, the variable is uninitialised and should be set to 0, and that fixes the problem on the affected machine.

Thanks to @nadiakempers for the report.

LourensVeen:
This fixes an uninitialised variable, which caused Huayno to crash in some cases. Fixes #1094.

### DIFF
--- a/src/amuse/community/huayno/src/evolve_sf.c
+++ b/src/amuse/community/huayno/src/evolve_sf.c
@@ -6,7 +6,7 @@
 static void drift_naive(int clevel,struct sys s, DOUBLE etime); /* drift/extrap sys to itime*/
 static void split(FLOAT dt, struct sys s, struct sys *slow, struct sys *fast);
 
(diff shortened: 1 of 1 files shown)


LourensVeen:
Argh, sorry, vim automatically fixed the whitespace because we have an editorconfig file now that tells it to do so. I'll clean this up, on sec.

LourensVeen:
There, you should able to see all glorious five characters now!

PR #1132, opened 2025-04-08
Move original units/constants and add updated ones

rieder:
AMUSE units are outdated: in some cases they contradict international agreed values.
In this PR units and constants are updated to their current standards.
As a reference, the default values used in Astropy should be used.

- [ ] Make original AMUSE units / constants available under a new name
- [ ] Update default AMUSE units / constants
- [ ] Update tests
- [ ] Test values of units against their Astropy equivalent
- [ ] Make sure that Astropy unit names map one-to-one to AMUSE units (create aliases where needed, maybe deprecate old AMUSE names)


rieder:
Not all units/constants have been updated yet, but the most important ones have.

PR #1135, opened 2025-04-24
Make issues stale on much longer timescales

rieder:


rieder:
@LourensVeen could you approve this one?

PR #1136, opened 2025-04-24
Update ci.yml - use more modern Python and Ubuntu

rieder:


rieder:
Python 3.9 is the oldest still-supported Python version - so bumping our CI to that version.

PR #1137, opened 2025-05-06
Replace the AMUSE build system

### LINKED ISSUE #1024: Build system needs replacing
# Issue

The current build system for AMUSE is based on setuptools' custom build commands feature, which is obsolete and in the process of being removed. To keep AMUSE from becoming impossible to build soon, we'll have to create a new build system. Additionally, I've been trying to build conda packages for AMUSE and keep running into issues with the build system there as well, so this should also be addressed.

The current build system comprises a global GNU autoconf-setup for detecting dependencies, a top-level Makefile which mostly just calls `setup.py`, which in turn uses a whole lot of custom Python code in `support/`. Community codes have an outer Makefile in `src/amuse/community/<code>/` and usually another build system (the original one that came with the code) inside a `src/` subdirectory. Some codes are downloaded at build time using a number of copies of a `download.py` script. Altogether, this builds the libraries in `lib/`, the community code workers, and installs the Python code for the framework and for the individual community codes.

Most of the actual work is done by the code in `support/`, which is rather complex and difficult to understand. Setuptools' inheritance-based architecture definitely doesn't help here either, but its deprecation means that that's about to be solved :smile:.

For all the above reasons, we need a new system.

# Usage scenarios

There seem to be a couple of different usage scenarios for AMUSE in general, all of which need to be supported well by the build system.

## Basic usage

In this scenario, the user installs AMUSE locally and uses it to run simulations without changing AMUSE itself. This typically occurs during organised courses, but also during student projects and even more advanced research projects that don't require large amounts of compute power. Often this is CPU-only, but if there is a GPU available then it should be usable. MPI may or may not be available.

For this scenario, AMUSE must be single-command installable, preferably using Conda as it is the most widely used suitable package manager, or by building from source if necessary. GNU/Linux, MacOS and WSL2 need to be supported and work out of the box (almost) every time.

## HPC usage

For larger simulations, an HPC machine may be required. More and more of these machines (and the workstations in Leiden too) use EasyBuild to install software, so an EasyBuild configuration should probably be provided for this, and we need a build system that works correctly in an EasyBuild environment. Despite its name, EasyBuild is not very easy to use, and for an end user with no experience with EasyBuild building from source is probably easier, although there is the dependency issue to contend with.

Conda may be an option here, as it has some support for using the local MPI, but getting this to work may be trickier than doing a source install. These environments pretty much all run Linux, and more and more often have GPUs that should be usable when available. Some community codes can support non-CUDA GPUs, and these are becoming more common, so that should work too.

## Community code development

In this scenario, the user installs AMUSE with the intent of using most of the system unchanged, except for a single community code which needs to be added, extended, improved or fixed. To be able to work efficiently, the user should need to run at most one command to recompile their code and make the new version available in a Python environment (virtualenv or conda), so that the changes can be tested quickly.

Note that the community code may be in the AMUSE repository or somewhere else (e.g. in OMUSE or in the upstream community code repository), but that the worker still needs to link with the libraries in the framework. This should also mix with package managers, so that you could e.g. create a conda environment, conda install amuse-framework into it, and with that active easily build and install the community code into the same environment.

# Requirements

The requirements seem to be:

- Build and install AMUSE into a Python environment (virtualenv or Conda)
  - built-in libraries (i.e. stopcond, forsockets, etc.),
    - also g6lib and sapporo_light for now (this should probably be a separate package entirely, and we should make sapporo2 the default, and maybe there should be a sapporo3 with a better API, but that's all for the future)
    - headers should be provided with the libraries, rather than with the code using them
    - we should look at static vs. dynamic linking here, static linking is kind of obsolete but dynamic linking requires ensuring that the libraries are found when starting the worker
  - Python framework and tools (amuse.lab, unit conversion, communication layer, amusifier, etc.),
  - community codes, individually and a single command that installs them all
    - We should have some options to build only a core set of community codes, or only codes that don't need a GPU, or everything.
    - Codes that aren't included should be downloaded automatically as needed.

- Works on GNU/Linux, Apple MacOS, and WSL2 (GNU/Windows)

- Automatically detect dependencies
  - Globally for the framework, and on a per-community-code basis
  - MPI optional for the framework
  - CUDA, OpenCL, Fortran, HDF5, NetCDF, etc. required for codes that require them

- Automatically download community codes as needed

- Can be used to build Conda binary packages
  - In particular, conda-forge builds MacOS binaries on Linux using a cross-compiler, and that needs to work. This may be an adventure for some of the community codes.
  - In general, it should adhere to all the usual conventions to make it easier to build packages for a variety of package managers

- When developing, the developer should be able to use a single command to do everything needed to rebuild a community code and make it available in an existing environment, so that they can quickly test the new version
  - Editable installs (pip and/or conda) should be used to avoid having to run anything at all if only the Python side of a community code is updated, and to avoid having to build a new package every time which especially for conda takes a long time to resolve dependencies

# Design

The new build system should be driven by a top-level Makefile, which should reduce a basic install to a single command. Make will call pip where needed, rather than the other way around. The central Makefile can be used to build community codes, either all, certain subsets suitable to the available hardware, or individual ones. Individual community codes can also be built and installed using a Makefile in their directory. The central Makefile calls the Makefiles in the community code directories to build them, just like the Python build system currently does. Targets for community code Makefiles are currently somewhat standardised, this should be cleaned up and tightened a bit. If different versions of a worker can be built (e.g. CPU and GPU, or CUDA and OpenCL), then detecting the environment and building the appropriate workers should be done inside the community code build system, not by the central Makefile.

Autoconf seems to be working okay at the moment, so it would be good to keep, but we'd like for the community codes to have their own autoconf setup that requires only the dependencies of that particular code (https://github.com/amusecode/amuse/issues/577, https://github.com/amusecode/amuse/issues/583). Currently the resulting `config.mk` is stored in amuse, and then retrieved later by the build system. This is causing no end of problems when building separate packages so it needs to go, but I'm not sure yet that that information is only used for building the community codes, or also for other things. To be investigated.

The worker binaries should not be built by the Python package build; instead they're built first by make, which will then call pip to do the install of the Python bits, the worker, and anything else needed (e.g. data). Community codes should have their own `pyproject.toml` in the `src/amuse/community/<code>/` directory, so that a Python package for only that code can be built from that directory. This also gets rid of the separate `packages/` directory. The framework will have its own separate `pyproject.toml`.

I think amusifier will generate some files for you to start with when adding a new community code. I need to look into this, but probably those templates will have to be updated to match the new system, and possibly some files will have to be added (configure.in, pyproject.toml). Needs investigation.

LourensVeen:
Well, it's been a very long time coming, but the to-do list is looking pretty empty and I guess that means that the new build system is done.

There are some broken codes still that need to be fixed, but those issues are unrelated to the build system. The only other items in #1024 are updating the development documentation and updating the CI. The development documentation looks pretty obsolete and needs a bigger clean-up that would be better to have as a separate issue. The CI is incomplete as-is, and I'd likewise like to do that separately as the next item on my list.

This does not merge cleanly, so I'll go fix that, but I wanted to make the PR at least.

Fixes #1024.


LourensVeen:
Just for my understanding, are your MESA commits fixes for something I broke, or improvements that are unrelated to the new build system? And what exactly gets fixed by adding MesaInterface to the imports? It's not clear from the commit message.

rieder:
> Just for my understanding, are your MESA commits fixes for something I broke, or improvements that are unrelated to the new build system?

A bit of both, the renaming of `MESA` -> `Mesa` (also for `Interface`) is unrelated but meant to standardize how classes are named. `MesaInterface` was not yet available, that was an omission that is now fixed (at the expense of `MESAInterface`). That should have been a separate PR probably.

It's mainly a fix of this PR though. In the pre-PR setup, both versions of MESA can be imported without causing an ImportError. My fixes restore that behaviour in the post-PR setup, by letting `from amuse.community.mesa import Mesa` fail only if no version of MESA is available.

> And what exactly gets fixed by adding MesaInterface to the imports?

A bit of a lazy fix, `amuse.community.mesa` imports both `Mesa` and `MesaInterface` and so both need to be available. Could also be fixed by skipping `MesaInterface` altogether, and that might be better.

> It's not clear from the commit message.

Sorry about that!

LourensVeen:
Okay, but we still have a bunch of instances of `MESAInterface` now, in the tests for MESA r15140 and r2208. And `class Mesa` still refers to `MESAInterface` causing the ext tests to fail (the code tests pass accidentally because they still import `MESAInterface`).

I'm just going to rename all remaining instances of `MESAInterface` to `MesaInterface` to fix that, and then we'll hope that this doesn't break anybody's scripts. Although most users probably never upgrade anyway.

rieder:
> Okay, but we still have a bunch of instances of `MESAInterface` now, in the tests for MESA r15140 and r2208. And `class Mesa` still refers to `MESAInterface` causing the ext tests to fail (the code tests pass accidentally because they still import `MESAInterface`).
> 
> I'm just going to rename all remaining instances of `MESAInterface` to `MesaInterface` to fix that, and then we'll hope that this doesn't break anybody's scripts. Although most users probably never upgrade anyway.

Thanks. `{Code}Interface` classes should never be used directly by end users in any case.

LourensVeen:
Thanks! That mocassin symbol should be imported from the package though, not duplicated in the framework. It would also really help if you could run the tests before committing and pushing, so that I don't have to review two commits instead of one...

rieder:
> Thanks! That mocassin symbol should be imported from the package though, not duplicated in the framework.

I wanted to replicate previous behaviour (not failing an example on an `ImportError` importing this unit if Mocassin is not installed), but I guess that makes sense now. A Mocassin using example would fail then in any case. Could you suggest alternative code?

> It would also really help if you could run the tests before committing and pushing, so that I don't have to review two commits instead of one...

Uhm, yes. Sorry.

rieder:
> Okay, but we still have a bunch of instances of `MESAInterface` now, in the tests for MESA r15140 and r2208. And `class Mesa` still refers to `MESAInterface` causing the ext tests to fail (the code tests pass accidentally because they still import `MESAInterface`).
> 
> I'm just going to rename all remaining instances of `MESAInterface` to `MesaInterface` to fix that, and then we'll hope that this doesn't break anybody's scripts. Although most users probably never upgrade anyway.

I needed this for something else so I already fixed this now (sorry if you were already working on it).

LourensVeen:
I was working on merging the main branch back into this one to make it mergeable, and did that, but I didn't push it yet because the tests were failing. So now they're hopefully fixed but things have diverged. Probably easiest to just redo the merge. Do you have anything else that you want to change?

rieder:
Not at this point. I'm having some trouble with the old Mesa but that can wait.

LourensVeen:
I had problems with 2208 as well actually, but when I did the exact same code in a Docker container with a venv it worked, so maybe there's an issue with the patching or something. AMUSE was complaining about the RPC protocol it seemed, but unfortunately the error message wasn't very helpful so I'm not sure what the cause is. I'm compiling your latest commits now both directly and in a Docker, and if that all works then I'll go and merge again.

LourensVeen:
Ah, great, it seems that `./setup test all` skipped running tests for the mesas, so it was probably broken inside the Docker container as well.

The problem there is that Python is just inconsistent, it specifies a way to normalise package (.conda or .whl) names that says all sequences of dash, underscore and period are replaced by a single dash, but then packages (directories with `__init__.py` in them) need to have underscores to be valid identifiers in Python. So there's no way to name things consistently. To make it worse, `pip list` shows underscores but `conda list` normalises all the names before printing, so the same installed package shows up as `amuse-mesa_r2208` or `amuse-mesa-r2208` depending on how you list it.

I had a work-around for that, but it turned out there was a corner case, so the new rule is that packages always use dashes (turning `amuse-mesa_r15140` into `amuse-mesa-r15140` and directories always underscores (e.g. `src/amuse_mesa_r15140`). This does mean that we now don't know from the name if `amuse-huayno-openmp` is the main package of `huayno-openmp` or a subpackage with an openmp worker for `huayno`, but we can check to see what we have a directory for and then it works.

If you have a package with an underscore currently installed and pull the changes, then `./setup uninstall amuse-mesa_r2208` doesn't work, and neither does `./setup uninstall amuse-mesa-r2208`. So you nede to `pip uninstall amuse-mesa_r2208` just once, and then you can use `./setup install amuse-mesa-r2208` to install it again with its new name.

I've also redone the merge and fixed a few more things that came up. MESA r2208 is still failing for me, so I need to figure that out still. And maybe update the CI to call the new build system so that it actually passes before we merge.

rieder (dismissed):

LourensVeen:
Ah, good point, I updated some of the stuff in `docs/` but I didn't update the README. Indeed, that command should get you started. I'll update the README and then I'm going to finally merge this thing :smile:

PR #1142, opened 2025-05-26
fix python2->3 problem and use new python raw strings to protect latex

teuben:
this is an example I can do in NEMO, so was curious how amuse did it.... except it fell victim to code rot


teuben:
I take it the new book examples will be in the amuse repo ?
Will we also get ride of the * import ?

PR #1146, opened 2025-05-27
Fix SimpleX

### LINKED ISSUE #1111: Simplex is broken
**Describe the bug**
Simplex doesn't compile.

This code has multiple issues:
- It includes a copy of HEALpix that is so obsolete it doesn't compile anymore
- It includes a copy of QHull that doesn't compile with clang
- It includes a copy of HDF5 that doesn't compile with clang
- It has obsolete source code itself and doesn't compile

The dependencies are all commonly available, so they should just be installed using a package manager, which will also keep them up to date. Simplex itself may need updates to work with newer versions of these, and the whole thing needs to be made to work with both gcc and clang.

LourensVeen:
SimpleX didn't build mostly because it vendored some ancient versions of QHull and HEALPix. This adds those dependencies to our configury so that we can get them from the system, removes them from SimpleX, converts SimpleX to the new build system and fixes it so that it works.

This currently works **only in a virtualenv and not in conda**, because of https://github.com/conda-forge/qhull-feedstock/issues/23. That needs to get fixed first, so that we can test this with conda too and then merge it.

@rieder could you give this a try on macOS in a venv? I have only tested it on Linux. Thanks!

Closes #1111.

### DIFF
--- a/src/amuse_simplex/support/configure
+++ b/src/amuse_simplex/support/configure
@@ -6039,12 +6039,12 @@ fi
 
     LIBS="$amuse_lib_qhull_save_libs"
 
-           for ac_header in libqhull/libqhull.h
+           for ac_header in libqhull_r/libqhull_r.h
 do :
-  ac_fn_cxx_check_header_compile "$LINENO" "libqhull/libqhull.h" "ac_cv_header_libqhull_libqhull_h" "$ac_includes_default"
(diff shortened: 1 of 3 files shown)


rieder:
Works with Conda, but not yet with venv (qhull is not detected as being installed).

rieder:
Still, recommended we merge this and treat that as a new, separate issue.

rieder (approved):

PR #1152, opened 2025-06-03
Implement the new CI

LourensVeen:
This doesn't yet implement anything, but this is a for so I need a PR to get the CI to actually run so that I can see what I'm doing. Work in progress.


LourensVeen:
Seems like the CI will also run on the fork, so closing this for now to reduce noise and CI minutes.

PR #1154, opened 2025-06-10
Update SeBa version

rieder:
Includes an important fix for SeBa - see https://github.com/amusecode/SeBa/issues/29


LourensVeen (approved):
Hard to see anything controversial in this. Nice find!

PR #1161, opened 2025-07-03
Include test particle calculations in Ph4

ErwanH29:
Added a couple of lines in Ph4 to allow for test particle computations. Stopping conditions neglect collisions between test particles. 
I closed the previous iteration to do a hard reset on my local repository, given the new AMUSE release.
The modified scripts (interface.cc and src/idata.cc) are no different to the previous pull request.


spzwart:
Looks good.

Issue #1158, opened 2025-06-24
Using new build system from v2025.5.0.rc1, receive errors when building amuse/ amuse-framework


bbanne:
When attempting to build the amuse packages in a new conda env (called amuse), I am receiving errors during the outlined process. I will include files with terminal and logs, but the summary is that I created and activated a new conda env, installed the dependencies as outlined in the build instructions and the output of the setup script, but when I go to run ./setup install all, I receive the following error: 

```
--------------------------------------------------------------------------
The Open MPI wrapper compiler was unable to find the specified compiler
x86_64-conda_cos6-linux-gnu-cc in your PATH.

Note that this compiler was either specified at configure time or in
one of several possible environment variables.
--------------------------------------------------------------------------
make[1]: *** [support/shared/lib-targets.mk:103: amuse_mpi.mo] Error 1
make[1]: Leaving directory '/opt/amuse/amuse-2025.5.0.rc1/lib/amuse_mpi'
make: *** [Makefile:17: amuse_mpi] Error 2
make: Leaving directory '/opt/amuse/amuse-2025.5.0.rc1/lib'
support/setup/installing.sh: line 98: ../support/logs/install-amuse-framework.exit_code: No such file or directory
cat: support/logs/install-amuse-framework.exit_code: No such file or directory

amuse-framework failed to build and install correctly.

This is not supposed to happen!

Please contact the developers on Slack or make an issue on GitHub,
so that we can fix this together.
```

I definitely have the prerequisites installed, and the env active, so if there was something else I needed to change as far en environment variables before beginning the installation, it wasn't included in either the docs or the setup script process. 

Our environment is rhel 8.10 with a conda version of 24.7.1. I've included output from the setup and install process, install logs, and the listing of all packages installed in our conda environment. Any additional guidance is appreciated. I am sure it is probably something I have missed.

[TerminalOutputAmuse.txt](https://github.com/user-attachments/files/20890414/TerminalOutputAmuse.txt)
[CondaEnvAmuse.txt](https://github.com/user-attachments/files/20890415/CondaEnvAmuse.txt)

LourensVeen:
Ah, apologies, yes it is but we had an issue building from tar archives.

What you've got there is actually not a clone, which is a copy of a git repository including git metadata, but an archive, which just contains the files stored in that repository without any git stuff. The AMUSE build system determines which version we're building from the git metadata, and an archive doesn't contain that, so it fails.

[...]

PR #1167, opened 2025-07-22
Fix more issues installing from a tarball

LourensVeen:
This addresses the issues raised in #1158, as well as a related issue where pkg-config files weren't created correctly.

### DIFF
--- a/src/amuse_simplex/packages/amuse-simplex/pyproject.toml
+++ b/src/amuse_simplex/packages/amuse-simplex/pyproject.toml
@@ -21,6 +21,9 @@ distance = "{next_version}.dev{distance}+{vcs}{rev}"
 dirty = "{base_version}+d{build_date:%Y%m%d}"
 distance-dirty = "{next_version}.dev{distance}+{vcs}{rev}.d{build_date:%Y%m%d}"
 
+[tool.versioningit.vcs]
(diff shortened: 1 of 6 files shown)


src/amuse_simplex/packages/amuse-simplex/pyproject.toml
```diff
@@ -21,6 +21,9 @@ distance = "{next_version}.dev{distance}+{vcs}{rev}"
 dirty = "{base_version}+d{build_date:%Y%m%d}"
 distance-dirty = "{next_version}.dev{distance}+{vcs}{rev}.d{build_date:%Y%m%d}"
 
+[tool.versioningit.vcs]
+method = { module = "version_helper", value = "get_amuse_version", module-dir = "amuse_simplex/support/shared" }
+
```
rieder, line 26: was this fix only needed specifically for Simplex?

rieder (approved):
One question, but I think this is fine.

src/amuse_simplex/packages/amuse-simplex/pyproject.toml
LourensVeen: Yeah, I converted the rest using a script so they were all good, then fixed Simplex (which had a bunch of issues) using the template I'd made previously for converting the codes, which didn't have these extra lines, so they were missing only there. I did check the rest to make sure :smile:.

PR #1179, opened 2025-10-01
Make arctan2 work with ScalarQuantities

rieder:
Ensures that x and y use the same unit, which is otherwise ignored (it is divided out).

### DIFF
--- a/src/amuse/units/trigo.py
+++ b/src/amuse/units/trigo.py
@@ -10,8 +10,8 @@ arcsin = lambda x: numpy.arcsin(x) | rad
 arccos = lambda x: numpy.arccos(x) | rad
 arctan = lambda x: numpy.arctan(x) | rad
 arctan2 = lambda x, y: numpy.arctan2(
-    x.value_in(x.unit),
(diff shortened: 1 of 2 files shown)


src/amuse/units/trigo.py
```diff
@@ -9,7 +9,10 @@
 arcsin = lambda x: numpy.arcsin(x) | rad
 arccos = lambda x: numpy.arccos(x) | rad
 arctan = lambda x: numpy.arctan(x) | rad
-arctan2 = lambda x, y: numpy.arctan2(x, y) | rad
+arctan2 = lambda x, y: numpy.arctan2(
+    x.value_in(x.unit) if isinstance(x, quantities.Quantity) else x,
+    y.value_in(x.unit) if isinstance(x, quantities.Quantity) else y,
```
rieder, line 14: Note that this explicitly checks the unit of `x` in both cases, and also whether `x` is a `Quantity`. This is because `x` and `y` need to be treated in the same way, otherwise they might have different units.

PR #1213, opened 2026-03-04
Fixes for issues 1212 and 1215

LourensVeen:
MESA r2208 and r15140 download a bunch of files on installation, from Zenodo, GitHub, and a university server. That university server does not like wget, and the official work-around is to set an empty User-Agent. So that's what we did, but recently Zenodo has started rejecting that with a 403, causing the MESA installs to be broken.

Additionally, the MESA people rearranged their GitHub repository recently, and since we were downloading from the master branch that broke our download, because the file is now somewhere else.

Finally, there was an issue with newer macOS (or XCode, not sure) not accepting two `-WL,rpath=` arguments on the command line, even if they add the same path. This tries to fix that as well. While messing around with this, I noticed some missing dependencies in the mesa-r15140 metadata, so I added those.

Bit of a mixed bag this, but we need to get this fixed also because we have a workshop coming up.

### DIFF
--- a/.github/workflows/test-framework.yml
+++ b/.github/workflows/test-framework.yml
@@ -39,7 +39,7 @@ jobs:
 
     - name: Install dependencies
       run: |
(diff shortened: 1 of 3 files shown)


rieder (commented):
It seems download speed (of mesa-r15140.zip from Zenodo) has become incredibly slow, could that be related to using `curl` instead of `wget`?

PR #1217, opened 2026-03-05
Try using MPICH on macOS to avoid crashes

LourensVeen:
The CI isn't working on macOS because we're getting an unknown MPI error. This switches to MPICH on macOS, see if that works...


LourensVeen:
Seems like it gets through `test_c_distributed_implementation` but now it's hanging on `test_fortran_implementation`. We need that CI machine in Leiden I think...

rieder (dismissed):
MPI on GitHub/macOS has been a mixed bag, it seems unpredictable whether it will work or not. If nowadays MPICH is a better option, fine!

Issue #1202, opened 2025-12-12
HDF5 detection broken?


LourensVeen:
**Describe the bug**
We're having trouble detecting HDF5 on some systems, in particular a cluster called rusty that is used by the Torch developers. It seems that the h5cc script is broken, and that that may be related to build system changes in HDF5. The package there is built using Nix, so it's also possible that that is doing something odd.

- [x] Investigate HDF5 build system changes
- [x] Create a bunch of from-source HDF5 builds with various options
- [ ] Try to detect them using the current AX_LIB_HDF5 macro
- [ ] Update or replace the macro with one that can detect all of them correctly
- [ ] Test on rusty as well, or maybe with a local Nix-built package

LourensVeen:
If I install `hdf5-helpers` on Ubuntu 24.04 but not `libhdf5-dev` or `libhdf5-mpi-dev`, then the macro we currently use will find `h5cc`, try and fail to compile and link, and then conclude that HDF5 is available.

LourensVeen:
I'm having some second thoughts on how this is implemented, with all the languages detected at the same time. It's more efficient, but it's not how the other macros work. Then again, maybe it's good enough. Let's wait for some testing from the Torch developers.

PR #1237, opened 2026-03-13
Add libz as an explicit dependency for SimpleX and MESA r15140

### LINKED ISSUE #1225: Simplex depends on zlib
**Describe the bug**
Simplex won't compile if -lz is not available, but this is not reflected in the metadata.

LourensVeen:
Turns out zlib is needed for HDF5, it's overlinked. With conda, installing HDF5 will automatically drag in libz, but with other package managers it's possible to install an hdf5 devel package but not a libz devel package, and then it's missing, so we need to test both.

MESA r15140 has the same issue, so it's added there as well.

There's one problem in that Fedora 40 has changed the name of the libz package because it's moving to zlib-ng, while other RPM based distributions still use the old one. I've left the old name for now, but we'll want to switch once the new name is more widespread.

Fixes #1225.


LourensVeen:
I've added a fix for a small issue I ran into detecting GSL and HDF5 without Conda. Note that HDF5 is still sometimes detected even though we don't have development packages; we'll fix that as part of #1202.

PR #1248, opened 2026-03-26
Mesa-r15140: ensure makedepf90 configure does not get overwritten

rieder:
On macOS, `make` tries to run `autoconf`, which causes building Mesa to fail. The reason is that the configure file is slightly out-of-date. This fixes that. One of the weirder bugs I've seen...


LourensVeen (approved):
Not the first time I've seen such an issue with MESA. All those nested tarballs...

PR #1251, opened 2026-03-26
Resolves #1175; resolves #1250: fix OpenMP detection on macOS with Homebrew libomp

### LINKED ISSUE #1175: OpenMP not detected when using XCode Clang
When running ./setup, openmp isn't detected when using a virtualenv and XCode Clang. We should figure out why and either improve the AX_OPENMP macro or write our own.

### LINKED ISSUE #1250: openmp is detected on macOS, but flags are not set
**Describe the bug**
OpenMP codes are compiled without OpenMP support on macOS

**To Reproduce**
- build AMUSE on a Mac
- install an OpenMP code (e.g. Fi)
- start a worker, e.g. with `sph = Fi(mode="openmp", redirection="none")`
- notice that it doesn't use OpenMP (e.g. cpu usage stays at 100% max, any OpenMP print statements aren't printed)

**Expected behavior**
The code uses OpenMP, and uses many cores.

**Environment (please complete the following information):**
 - OS and version: macOS Tahoe, Conda-installed AMUSE

**Additional context**
Probably good to add some reporting to Fi, e.g. number of threads available?

kaiwu-astro:
This fixes OpenMP detection on macOS when using Apple Clang with Homebrew's `libomp`.

  AMUSE's current OpenMP detection uses `AX_OPENMP`, but on macOS that was not enough:
  - `libomp` is keg-only in Homebrew, so its include/library paths are not found automatically
  - Apple Clang does not accept plain `-fopenmp` the same way GCC does

  As a result, `./setup` reported `openmp` as missing even when `brew install libomp` had already been
  done.

  ## Changes

  - update `support/shared/m4/ax_openmp.m4`
  - on `darwin*`, detect Homebrew `libomp` and add its include/lib paths
  - extend OpenMP flag probing to include Apple Clang compatible flag combinations
  - then I ran `support/update_configure_scripts` to regenerate the affected `configure` and `aclocal.m4` files across the repository (sorry this makes this PR look horrible, with 30k plus and 24k minus...)

  ## Result

  On this macOS machine with:
  - Apple clang 17.0.0
  - Homebrew `libomp` installed

  `./setup` now reports `openmp` in detected features, and the `** Disabled packages **` section no
  longer contains entries with `missing features: openmp`.


rieder:
I don't think this is the right solution (or at least not the complete solution). I don't have Homebrew's `libomp`, yet manually setting `OPENMP_FFLAGS = -fopenmp` gave the desired result for `Fi` (that's of course a Fortran code, so it's not using `clang` but `gfortran`...). 
I would also prefer not to be dependant on Homebrew now that we install all prerequisites through Conda.

PR #1252, opened 2026-03-31
remove pkg_resources as it is deprecated and not used

### LINKED ISSUE #1155: pkg_resources is deprecated
We seem to be using pkg_resources somewhere for something, and we've been getting warnings about it being deprecated for several years, but it seems like there's a deadline now:

> ../../../../../miniconda3/envs/test/lib/python3.12/site-packages/amuse/support/options.py:11
>   /Users/runner/miniconda3/envs/test/lib/python3.12/site-packages/amuse/support/options.py:11: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
>     import pkg_resources

We're now using hatchling instead of setuptools, although I'm not sure if that's relevant. At any rate, we should investigate what this is used for, whether we need that still, and if so, what the new way of doing that is.

rieder:
Fixes #1155. It is safe to remove as we don't include an `amuserc` file in this location in any case.

### DIFF
--- a/src/amuse/support/options.py
+++ b/src/amuse/support/options.py
@@ -7,11 +7,6 @@ from io import StringIO
 from amuse.support.core import late
 from amuse.support import exceptions
 
-try:
-    import pkg_resources
-except ImportError:
-    pkg_resources = None
-
 
 class GlobalOptions:
     INSTANCE = None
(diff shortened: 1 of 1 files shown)


LourensVeen (approved):
Yep, let's put this in and get rid of that warning/error.

Issue #1144, opened 2025-05-27
New Continuous Integration


LourensVeen:
**Is your feature request related to a problem? Please describe.**
The old CI wasn't working too well, and didn't cover all codes. It's been updated a bit to work with the new build system, but it's still hanging on Linux and it has MPI issues from time to time. We really need a reliable CI to pull us out of the current broken windows situation.

**Describe the solution you'd like**
A new CI setup with:
- Checks on every PR that run the framework tests and tests for any codes that have changed
- Regular builds on a variety of systems which build everything and run the tests
- Something with a GPU so that we can run tests for the GPU codes as well

**Describe alternatives you've considered**
We're currently having MPI issues with the free GitHub Actions runners, and they don't have GPU support. We could consider going with paid GitHub runners, an external service ([Ubicloud](https://www.ubicloud.com) seems like a good option), or a dedicated physical machine in Leiden with a GPU in it. A service would have the advantage that you don't need to maintain it and it's more energy efficient, a local machine would be more flexible and allow easier local experimentation. Cost would have to be considered too.

LourensVeen:
Update on this: I've managed to make the GitHub CI a bit more reliable (see #1260), and we have a local machine at the Sterrewacht with 4 cores, 16GB RAM and a GPU.

I'll finish the workflows first, then try to connect that machine to GitHub (and BitBucket, for Torch, but that's another story).

Issue #1259, opened 2026-04-23
amuse-mesa-r15140 build fails


amitavdas:
**Bug**
I was trying to install `amuse-mesa-r15140`. But the output during install said: "This is not supposed to happen!"
I think build failed. I am reporting this only cause the message suggested to do so.

**Reproducing**
`./setup develop amuse-mesa-r15140`

**Logs**
Logfile to help explain the problem.

[develop-amuse-mesa-r15140.log](https://github.com/user-attachments/files/27020710/develop-amuse-mesa-r15140.log)

**Environment:**
 - OS and version: Arch Linux `6.18.22-1-lts`
 - Compiler: ```gcc version 15.2.1 20260209 (GCC)```

LourensVeen:
Hi Amitav,

It's difficult for us to test every single combination of compiler and library and system environment, so we try our best but sometimes have to rely on users to report any problems. That's why that message is there, and thank you for following up on it :pray:.

It seems that the problem here is that GCC 15 sets the default C standard to C23, and MESA r15140 contains code that isn't compatible with that. Since MESA also doesn't tell the compiler which C standard it wants, the compilation fails.

I'm going to try to reproduce this locally and see if I can add another tweak to the MESA configuration to fix this. Would you be willing to try to compile from a git branch to test?

LourensVeen:
I've managed to reproduce this on Ubuntu 26.04 in a virtualenv, with GCC 15.2.0. Some other codes also fail to build, for which I'll make separate issues.

This one seems to actually be in the build of makedepf90, which contains an enum to define `true` and `false`, which have become keywords in C23. Compiling it with `-std=gnu11` works, but not `-std=c11` because that doesn't have `strcasecmp`. Not sure why, because it's POSIX 2008, but there we have it. So adding that flag to the makedepf90 build seems to fix it.

I've made a branch with that fix, and it compiles, but the tests fail, and I haven't been able to figure out why yet. To be continued.

LourensVeen:
It's hard to figure out what's going wrong. MESA's solver doesn't converge on the failed tests, and the only difference between working and non-working builds I've found so far is that the working ones use GCC 13.4.0, while the non-working ones use GCC 14.3.0 or GCC 15.2.0. The GCC changelog doesn't suggest anything either.

Issue #1163, opened 2025-07-11
SeBa, MESA r2208 and GalactICS fail precision tests on ARM processors


LourensVeen:
When running on ARM processors (Apple Silicon and also Raspberry Pi), SeBa, MESA r2208 and GalactICS fail some of the tests because of what seems to be a numerical precision issue. GitHub Actions runs:

GalactICS: https://github.com/LourensVeen/amuse/actions/runs/15856287284
SeBa: https://github.com/LourensVeen/amuse/actions/runs/15846565565
MESA r2208: https://github.com/LourensVeen/amuse/actions/runs/15846565519

These work on my AMD Zen laptop with Linux and both GCC and Clang, and on GitHub actions (Intel?) with gcc. Also SeBa is in C++, and GalactICS and MESA are in Fortran, so it doesn't seem to be a compiler or OS issue.

The one thing we can think of at the moment is that it might be the weird 80-bit internal long double that x86 FPUs have and ARM FPUs don't. There's a compiler switch to use only SSE on x86, which doesn't do more than 64 bits, so we can try that to see if that triggers the problem on x86 as well.

LourensVeen:
I'm working on the CI, and noticed that I had two different Linux runs, one where MESA r2208 fails the precision test and one where it passes it. The difference seems to be in the gfortran version: 13.4.0 passes, 14.3.0 fails. We're also having convergence problems with MESA r15140 on 14 and up (see #1259).

Pass: https://github.com/amusecode/amuse/actions/runs/26024686293/job/76494861614
Fail: https://github.com/amusecode/amuse/actions/runs/26024686132/job/76494860603

LourensVeen:
Possibly similar: https://github.com/Reference-LAPACK/lapack/issues/1073, which suggests trying `-fno-expensive-optimizations` and `-fno-tree-loop-vectorize`.

That references https://gcc.gnu.org/bugzilla/show_bug.cgi?id=120817, which fixes the compiler issue, but the only release since then has been GCC 15.2.

LourensVeen:
I can reproduce this for SeBa on Ubuntu with conda and GCC 14.3.0, and the above options don't help. Turning off optimisation completely with `-O0` doesn't help either.

LourensVeen:
[...]

PR #1260, opened 2026-04-28
Issue 1144 new continuous integration

LourensVeen:
This adds a whole bunch of CI workflows, one for each code, which run whenever that workflow is changed or when the code is updated. This also contains some technical fixes for various codes, where tests that are too resource intensive for the CI are disabled.

Some of the codes remain broken and need attention from an expert, but I think it's better to have a CI that tells you that things are in that state, than to not have one at all, so after I do some more cleanup I think we should merge this anyway and then go look for someone who can fix things.

Addresses #1144


LourensVeen:
It turns out that getting working MPI and compilers with Homebrew and MacPorts isn't easy. I'd like to get this merged, so I'm going to remove those tests for the time being, so that we at least have per-code tests on Ubuntu and macOS, plus an all-up conda based install and test.

LourensVeen:
Okay, I think this is ready to go now. Some of the codes are still failing consistently:

- evtwin on macOS with a precision issue (related to #1163?)
- MESA r2208 on macOS (#1163)
- MESA r15140 because #1248 broke its build system (in work for #1259)
- PeTar on macOS because of bad C++ (see #1264)
- SeBa with a precision issue (#1163)

There are also still intermittent timeouts and infrastructure failures, probably because these tests are quite heavy, use MPI, and because this PR runs all of them simultaneously. It's better than it was though due to additional OpenMPI configuration and a switch to MPICH on the macOS side. In normal circumstances code-specific tests will run only for PRs touching that code.

This does not include weekly regression tests for macOS with Homebrew and MacPorts, because I had trouble setting those up correctly. You don't just install a compiler or an MPI library with these things and have a working setup. It would help if someone with more mac expertise could do a nice clean setup for that.

Let's merge this and continue to set up our custom runner, and also to fix the codes so that everything passes all the time and we don't have broken windows anymore.

rieder:
Ok for me - not sure how #1248 broke MESA but this is not the right place to discuss about that

LourensVeen:
Make doesn't work if you use spaces instead of tabs :smile:

rieder (commented):
There seem to be a lot of non-CI changes (e.g. to Makefiles) in this PR, is that by design or by accident?

rieder:
> Make doesn't work if you use spaces instead of tabs 😄

🤦
(that's easy to fix though)

LourensVeen:
By design, it includes some fixes for issues I ran into:

- Make `aarsethzare`, `fractalcluster`, `galactics` and `krome` use the Fortran compiler detected by autoconf, rather than its default, which doesn't always work
- Use LDFLAGS when linking `adaptb`'s executable, to ensure libs are found correctly
- Skip overly expensive tests that don't run (well) on the CI in `gadget2`, `hermite`, `mesa-r15140`, `petar`, and the framework tests.
- Specify precision in `galactics` test, IIRC the default changed after this test was written
- Compile `hermite_grx` as C++11 explicitly to avoid compiler errors
- Fix C++ compile issue in `mi6`, `register` keyword is obsolete
- Use `python -m pip install` instead of `pip install` in `tupan` build for better reliability (and to make it work on the CI)
- Fix resource leaks and other small issue in the framework tests
- Fix automatic installation of `mpi4py` in a venv on macOS

and somehow a bunch of whitespace fixes in concurrent.py snuck in, sorry.

rieder (dismissed):

rieder:
Looks good

LourensVeen:
Okay, there are only known issues left in the codes, which this is supposed to detect, and everyone agrees that it's a good idea, so I'm going to merge this so that we have a working CI, finally.

Issue #1209, opened 2026-01-27
Check if Sapporo 2 works with GPUs


HannoSpreeuw:
Does Sapporo 2 work with NVIDIA GPUs

HannoSpreeuw:
I managed to build Sapporo 2 as "included" in AMUSE with CUDA release 13.1, with a few changes that have to do with newer CUDA versions.

@rieder I could file a MR to https://github.com/treecode/sapporo2 with those updates.
After the merge a Sapporo 2 build within AMUSE should be as simple as
```
cd lib/sapporo_2
python download.py
cd src/sapporo2-master/lib/
make
```

The build is successful
```
ar: creating libsapporo.a
a - sapporohostclass.o
a - interfaces/sapporoG6lib.o
a - interfaces/sapporoYeblib.o
a - interfaces/sapporoG5lib.o
``` 
but yields a ton of warnings such as
```
./include/cudadev.h:357:45: warning: dereferencing type-punned pointer might 
break strict-aliasing rules [-Wstrict-aliasing]
```
and
```
./include/sapdevclass.h:103:19: warning: 'double4' is deprecated: 
use double4_16a or double4_32a [-Wdeprecated-declarations]
```

HannoSpreeuw:
I guess I will need to try running as well.

LourensVeen:
I think `/lib/sapporo2` still needs to be integrated in the new build system; I guess that's a job for me.

And then we need to figure out whether the codes work with Sapporo 2. I think there's one that requires Sapporo 2 and doesn't work with Sapporo light, but I don't remember which.

HannoSpreeuw:
I'd be happy with a Python script that works for both Sapporo Light and Sapporo 2, with some switch.

I guess that switch requires a manual modification of the Makefile that steers `./setup install amuse-ph4-sapporo`.

LourensVeen:
Actually, maybe `./setup install sapporo2` and `./setup install amuse-ph4-sapporo` and `./setup install amuse-ph4-sapporo2`? That would be easiest to implement I think.

HannoSpreeuw:
> That would be easiest to implement I think.

And which files would this implementation affect?

LourensVeen:
[...]

PR #1262, opened 2026-05-11
Sapporo2 build system integration

LourensVeen:
This adds Sapporo2 to the build system. We should merge https://github.com/treecode/sapporo2/pull/16 first, and then this needs to be updated with the version hash of the new commit created by that merge.

Note that there's a commit in here that fixes the autoconf name for sapporo_light. That's unrelated and it makes absolutely no difference to anything functionally, but I discovered this copy-pasted error and decided to fix it anyway for hair splitting purposes.

Partially addresses https://github.com/amusecode/amuse/issues/1209


lib/sapporo_2/support/configure
HannoSpreeuw, line 1: Why is this such a huge file?

lib/sapporo_2/Makefile
```diff
@@ -1,10 +1,92 @@
...
 
-.PHONY:
+
+# Main targets
+.PHONY: all
+all: sapporo2
+
+.DEFAULT_GOAL := all
+
+VERSION := eeac0835f80d6e0e8ca778bdea4dbf3ae799ca29
```
HannoSpreeuw: VERSION := 7cae51ed1368ea457fb5f3ca01b4775f7558a9e7

HannoSpreeuw (dismissed):

lib/sapporo_2/Makefile
HannoSpreeuw: I will add this.

HannoSpreeuw: Done through commit [65164fa](https://github.com/amusecode/amuse/pull/1262/commits/65164fa761d4f7b71ad9d085a416835b24100322).

lib/sapporo_2/support/configure
LourensVeen: It is generated using a complicated set of nested macros. It's not intended for human consumption, so as long as it works that's fine. It's `configure.ac` that is the input for this, so if that looks okay then it's fine.

HannoSpreeuw: `lib/sapporo_2/support/configure.ac` looks fine to me.

LourensVeen:
I've removed the `-sapporo` package in favour of `-sapporo2`. I think the only thing we need now is the `make test` target for Sapporo2, so that `./setup test sapporo2` will work?

HannoSpreeuw:
> I think the only thing we need now is the `make test` target for Sapporo2, so that `./setup test sapporo2` will work?

Yes, see [Sapporo 2 PR 20](https://github.com/treecode/sapporo2/pull/20).

PR #1266, opened 2026-05-19
1207 fix

### LINKED ISSUE #1207: Ph4 gpu capable in handling test particles
It would be nice if the gpu version of Ph4 can handle test particles.

**Describe the solution you’d like**
Support for test particles in the GPU version of Ph4, such that massless particles are excluded from the source (j) interaction loop while still allowing them to receive accelerations as targets (i).

**Describe alternatives you've considered**
I've tried adding a mass condition within the `dev_evaluate_gravity.cu` file akin to that in the CPU version of Ph4, but no speed-up is found. As a result, I think most  of the heavy work goes into synchronisation of states or building the arrays. I could look into it, but given my lack of knowledge revolving gpu codes, I would be very happy if someone with more experience could tackle it.

HannoSpreeuw:
Fixes #1207 at the level of PH4, i.e. at the level of jdata.cc

[Massless particles will no longer be added as j-particles.](https://github.com/HannoSpreeuw/amuse/blob/1d0386dffb7059a97ed0be25a1cf4328ed1ac760/src/amuse_ph4/src/jdata.cc#L143-L148)
This will work for both CPU and GPU backends in PH4.

Consequently, [the conditional in line 208 in idata.cc](https://github.com/HannoSpreeuw/amuse/blob/1d0386dffb7059a97ed0be25a1cf4328ed1ac760/src/amuse_ph4/src/idata.cc#L207-208) 
```
		// Skip particles with zero mass.
		if (jdat->mass[j] > _TINY_){
```		
will become redundant.


lib/sapporo_light/sapporo.h
```diff
@@ -22,6 +22,15 @@
 #include <map>
 
 #include <math.h>
+
+#ifndef SAPPORO_TEST_PARTICLE_MASS
+# if defined(_TINY_)
```
LourensVeen, line 27: Is _TINY_ ever not defined? If so, why?

src/amuse_ph4/src/jdata.cc
```diff
@@ -139,6 +139,14 @@ int jdata::add_particle(real pmass, real pradius,
 			real dt)		// default = -1
 {
     const char *in_function = "jdata::add_particle";
+
+    if (pmass <= _TINY_) {
+        if (mpi_rank == 0) {
+            cerr << "Warning: ignoring particle with mass="
```
LourensVeen: Is this going to produce a line of output for every massless test particle? That could be a lot, and what does that add? After all, ignoring a particle with no mass won't make a difference to the result anyway.

LourensVeen (commented):
Thanks! I have some questions....

lib/sapporo_light/sapporo.h
HannoSpreeuw: `_TINY_` may not be defined if PH4 is not the running AMUSE code in which case the [`_TINY_` definition in stdinc.h](https://github.com/HannoSpreeuw/amuse/blob/1d0386dffb7059a97ed0be25a1cf4328ed1ac760/src/amuse_ph4/src/stdinc.h#L33) may be skipped:
```
#define _TINY_	   (pow(2.0, -52))	  // double limit on powers of 2
```

src/amuse_ph4/src/jdata.cc
HannoSpreeuw: Good point, let me remove that.

HannoSpreeuw: Addressed through commit [3a2a913](https://github.com/amusecode/amuse/pull/1266/commits/3a2a91331a68a7f4e6583fab8e8022a8d6e03e3b).

ErwanH29:
I've tested this on one of my clusters, integrating until 100 yrs and testing N = 1e3, 1e4, 1e5, 1e6 test particles. it seems to scale linearly, so should be working!

HannoSpreeuw:
> I've tested this on one of my clusters, integrating until 100 yrs and testing N = 1e3, 1e4, 1e5, 1e6 test particles. it seems to scale linearly, so should be working!

@LourensVeen Do you think we can merge?

LourensVeen:
If it works for you and for Erwan, then I'd say it's ready to go. The failing test is an infrastructure problem, and anyway it tests only the framework, not Ph4, let alone on a GPU, so it doesn't say anything about whether this change is good or not. (That should be fixed soon, but that's a different PR.) So let's go!

PR #1270, opened 2026-05-29
Adding Tidymess to AMUSE documentation

elkogerville:
This PR adds a section to the AMUSE documentation on how to add a C++ code to AMUSE, using Tidymess as an example.

The documentation is still a work in progress, with many sections still needed. This PR is intended for feedback rather than merging into the codebase.

### DIFF
--- a/doc/tutorial/adding_example_code_tidymess.rst
+++ b/doc/tutorial/adding_example_code_tidymess.rst
@@ -101,3 +101,45 @@ Setting up the Makefile
 ~~~~~~~~~~~~~~~~~~~~~~~
 With our build system detection working, we now need to download ``TIDYMESS`` into ``AMUSE``!
 This is where the Makefiles come in. ``AMUSE`` packages typically have 2 Makefiles.
+
+
+Creating the Interfaces
+=======================
+With ``TIDYMESS`` compiled into ``AMUSE``, we can now begin the process of creating our interface!
+The interface system allows community codes, which are all unique and depend on diverse libraries and
+programming languages, to communicate with the ``AMUSE`` framework, which is native Python. ``AMUSE``
+interfaces define a number of interface functions, which provide a standardized way for ``AMUSE`` to
+communicate with each community code. This way, the experience of using any ``AMUSE`` code is identical:
+all codes can be evolved with the ``evolve_model`` method, particles are represented as
+``amuse.datamodel.particles``, etc... The strength of ``AMUSE`` lies in its ability to prototype quickly:
+If the user wants to see what solution a different code would give for the same calculation, all they have
+to do is switch which code they are using, and the script most likely does not need to change.
+
+Therefore, our job when creating an interface is to map the community code functions to the ``AMUSE`` interface
+functions. The amusifier already created all the files we need: the ``interface.py`` and the ``interface.cc``.
+
+Defining the Python interface
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The ``interface.py`` actually defines two classes: the high-level and low-level Python interfaces. The high-level
+interface is what the user interacts with when using the community code. This defines the methods, parameters,
+and properties of each community code, and defines how the ``amuse.datamodel.particles`` work within that code.
+The ``amusifer`` generates a minimal code stub for the ``interface.py``, which is enough to get us started.
+The process of defining our interface starts with figuring out what type of code we are adding to ``AMUSE``.
+``AMUSE`` has a set of predefined Python interfaces we can use to build our interface from.
+
+
++----------------------------------+-------------------------------------------+
+| Interface:                       | Example codes:                            |
++----------------------------------+-------------------------------------------+
+| ``GravitationDynamicsInterface`` | N-body: Tidymess, Ph4, Huyano             |
++----------------------------------+-------------------------------------------+
+| ``HydrodynamicsInterface``       | Hydrodynamical: Capreole                  |
++----------------------------------+-------------------------------------------+
+| ``MagnetoHydrodynamicsInterface``| MHD: Athena                               |
++----------------------------------+-------------------------------------------+
+| ``StellarEvolutionInterface``    | Stellar: MESA, EVtwin, SeBa               |
++----------------------------------+-------------------------------------------+
+
+These interfaces define many
+of the required interface functions so that we don't have to.
+In this case, ``TIDYMESS`` is a N-Body code, and therefore falls under the
--- a/doc/tutorial/adding_example_code_tidymess.rst
+++ b/doc/tutorial/adding_example_code_tidymess.rst
@@ -20,8 +20,8 @@ Getting Started
 ===============
 
 This tutorial assumes you have a working amuse or amuse development build,
-preferrably in seperated environment (virtualenv, venv or conda etc).
-Please ensure that amuse is setup correctly, this can be verified by running the
+preferably in seperated environment (virtualenv, venv or conda etc).
+Please ensure that amuse is set up correctly, this can be verified by running the
 ``amusifier`` .
 
 .. code-block:: bash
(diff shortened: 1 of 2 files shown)


doc/tutorial/adding_example_code_tidymess.rst
```diff
@@ -0,0 +1,145 @@
+.. _adding_example_code_tidymess:
+
+Adding a C++ N-Body Code to AMUSE
+=================================
+
+In this tutorial, we will create an interface from scratch for the
+TIdal DYnamics of Multi-body ExtraSolar Systems code, or ``TIDYMESS``,
+written by Dr. Tjarda Boekholt and Dr. Alexandre Correia. This code
```
LourensVeen: Is there a paper we can cite here?

doc/tutorial/adding_example_code_tidymess.rst
```diff
@@ -0,0 +1,145 @@
...
+
+    In this guide, ``TIDYMESS`` refers to the standalone simulation package,
+    while ``Tidymess`` refers to the ``TIDYMESS`` package inside of ``AMUSE``.
+
+
+Getting Started
+===============
+
+This tutorial assumes you have a working amuse or amuse development build,
+preferrably in seperated environment (virtualenv, venv or conda etc).
```
LourensVeen: preferably

doc/tutorial/adding_example_code_tidymess.rst
```diff
@@ -0,0 +1,145 @@
...
+Please ensure that amuse is setup correctly, this can be verified by running the
```
LourensVeen: set up

doc/tutorial/adding_example_code_tidymess.rst
```diff
@@ -0,0 +1,145 @@
...
+
+Since ``TIDYMESS`` is a native C++ code with no other dependencies, we will specify
+``--type=c``, but the ``amusifier`` can also build the interface directory for
+``f90`` and ``python`` codes.
+
+.. code-block:: bash
+
+    > amusifier --type=c --mode=dir Tidymess
+
+Having run the ``amusifier``, we now have our new directory in ``amuse/src/amuse_tidymess``.
```
LourensVeen, line 51: Well, that depends. Did we move into `amuse/src` first?

doc/tutorial/adding_example_code_tidymess.rst
```diff
@@ -0,0 +1,145 @@
...
+There should be all the required folders for building our interface, as well as a few code
+stubs to expand upon.
+
+Building the code
+=================
+Before we start working on the interface, we should try and install and compile ``TIDYMESS``
```
LourensVeen: "try to install"

doc/tutorial/adding_example_code_tidymess.rst
```diff
@@ -0,0 +1,145 @@
...
+
+.. code-block:: text
+
+   c c++ fortran java python cmake install download mpi openmp cuda opencl x11 opengl blas lapack gsl gmp mpfr fftw hdf5 netcdf4
+
+Since ``TIDYMESS`` is a standalone C++ code, we can delete most of those and simplify our dependencies to:
+
+.. code-block:: text
+
+   c c++
```
LourensVeen: Do we actually need `c`? The code is C++, and the bindings too, right? Or am I missing something again?

doc/tutorial/adding_example_code_tidymess.rst
```diff
@@ -0,0 +1,145 @@
...
+
+Setting up Autoconf
+~~~~~~~~~~~~~~~~~~~
+The ``amuse_deps`` file we just created informs the ``AMUSE`` build system about whether or not our
```
LourensVeen: I would remove this first sentence, as it's repetitive and not very related to the rest of this paragraph. Instead, maybe add an instruction to the previous paragraph for the user to go to the top-level directory and run `./setup`, so that they can see their code in the list?

doc/tutorial/adding_example_code_tidymess.rst
```diff
@@ -0,0 +1,145 @@
...
+of all the compiler and library variables needed for our package.
+
+Once these files are cleaned up, run ``autoreconf`` to (re)create the ``configure`` script, then run
+``./configure``. This will test the detection and check for errors. As a sanity check, run ``cat config.mk``
+and ensure that there are no ``@VARIABLE@`` symbols left! If there are, check that ``configure.ac`` and
+``config.mk.in`` were setup correctly.
+
+.. WARNING::
+
+    Make sure the the ``amuse_tidymess/support/shared/`` folder is a simlink to ``amuse/support/shared/``
```
LourensVeen: "symlink", it's a symbolic link not a simulated one. Although that would be appropriate in this context :smile:

doc/tutorial/adding_example_code_tidymess.rst
```diff
@@ -0,0 +1,145 @@
...
+Please ensure that amuse is setup correctly, this can be verified by running the
+``amusifier`` .
+
+.. code-block:: bash
+
+    > amusifier --help
+
+Naming our project
+~~~~~~~~~~~~~~~~~~
+Amuse naming conventions typically follows PascalCase, so we will name our project Tidymess.
```
LourensVeen: It's a class name, and [PEP-8 says they should be formatted like that](https://peps.python.org/pep-0008/#class-names). So nothing AMUSE-specific.

doc/tutorial/adding_example_code_tidymess.rst
LourensVeen: Also, there are two options: inside AMUSE here, and stand-alone in any directory or a separate repository or whatever. But I think it would be best to explain the stand-alone option in a separate section at the end.

doc/tutorial/adding_example_code_tidymess.rst
```diff
@@ -0,0 +1,145 @@
...
+and properties of each community code, and defines how the ``amuse.datamodel.particles`` work within that code.
+The ``amusifer`` generates a minimal code stub for the ``interface.py``, which is enough to get us started.
+The process of defining our interface starts with figuring out what type of code we are adding to ``AMUSE``.
+``AMUSE`` has a set of predefined Python interfaces we can use to build our interface from.
+
+
++----------------------------------+-------------------------------------------+
+| Interface:                       | Example codes:                            |
++----------------------------------+-------------------------------------------+
+| ``GravitationDynamicsInterface`` | N-body: Tidymess, Ph4, Huyano             |
```
LourensVeen: "Huayno", it's an Andean dance.

doc/tutorial/adding_example_code_tidymess.rst
```diff
@@ -0,0 +1,145 @@
...
+``amuse.datamodel.particles``, etc... The strength of ``AMUSE`` lies in its ability to prototype quickly:
+If the user wants to see what solution a different code would give for the same calculation, all they have
+to do is switch which code they are using, and the script most likely does not need to change.
+
+Therefore, our job when creating an interface is to map the community code functions to the ``AMUSE`` interface
+functions. The amusifier already created all the files we need: the ``interface.py`` and the ``interface.cc``.
+
+Defining the Python interface
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The ``interface.py`` actually defines two classes: the high-level and low-level Python interfaces. The high-level
```
LourensVeen, line 336: This paragraph could use some work. Maybe instead of "high-level" and "low-level" we could call it the Python interface and the native interface? That's a bit more descriptive (although that term native may also not be well-known, hmm). It's not explained what the low-level interface does anyway.

As for the structure, this mixes explanation of concepts, a practical description of what the amusifier does, and the next step to do. I think it's better to 1) briefly introduce the structure (two interfaces, user calls Python interface, which is handled by AMUSE and results in a call to the native interface, which calls the code or accesses data), then 2) in the next paragraph explain that the first step is making the Python interface, introduce the concept of standard interfaces, and then 3) we get to the practical side and we can mention that there's a stub for `interface.py` already there.

LourensVeen (commented):
This is definitely going in the right direction! I've added some comments with more improvements, but go ahead and focus on adding more content if you want, they can be fixed later as well.

PR #1272, opened 2026-06-01
Improve framework CI running

### LINKED ISSUE #1271: Framework CI should run for changes in /lib and tests
**Describe the bug**
The amuse-framework CI workflow currently only runs for changes in `src/amuse`, but the libraries are also part of the framework, and changes to the framework and ext tests should also trigger it.

**To Reproduce**
Make a PR that modifies something in `/lib`

**Expected behavior**
The `amuse-framework` test to run.

**Logs**
See e.g. https://github.com/amusecode/amuse/pull/1262

**Environment (please complete the following information):**
 - GitHub Actions

LourensVeen:
This extends the framework CI tests to run if there are changes to the libraries or the framework tests, rather than only to the Python code.

It also runs the all-up compat tests for PRs only if the PR changes the installer, otherwise only the specific tests for framework or code will be run.

Fixes #1271

### DIFF
--- a/.github/workflows/test-compat-conda.yml
+++ b/.github/workflows/test-compat-conda.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches:
       - main
+    paths:
+      - setup
+      - support/**
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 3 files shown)


LourensVeen:
Seems to work, and will save a bunch of email, so I'm going to merge this.

